PR #2, opened 2026-03-17
Add inkbox-browser-use example

dimavrem22:
## Summary

- Adds `examples/python/browser-use/` — packages the Browser Use + Inkbox demo following the same structure as the kernel example (CLI, config, identity selector, agent loop, email tools on a Controller)
- Supports `--env local|cloud`, `--chrome-debug-url`, and `--keep-browser` CLI flags
- Auto-launches Chrome in debug mode when using local browser if not already running
- Fixes identity selector in both kernel and browser-use examples to fetch full identity data (showing actual mailbox email instead of "no mailbox")


examples/python/browser-use/src/agent.py
```diff
@@ -0,0 +1,152 @@
...
+from src.config import Config
+from src.tools import build_controller
+
+if TYPE_CHECKING:
+    from inkbox import Inkbox
+    from inkbox.agent_identity import AgentIdentity
+
+logger = logging.getLogger(__name__)
+
+SYSTEM_PROMPT = """\
```
alex-w-99: Can we please have this file read in and use the associated `SKILL.md` file, a la `examples/python/kernel/src/agent.py`?

dimavrem22: fixed

alex-w-99 (approved):

PR #30, opened 2026-04-03
Agent self-signup

rayruizhiliao:
## Summary

- Add agent self-signup support to the Python SDK, TypeScript SDK, and CLI — matching the server-side endpoints shipped in inkbox-ai/servers#34
- Python SDK: 4 new classmethods on `Inkbox` (`signup`, `verify_signup`, `resend_signup_verification`, `get_signup_status`) + new `agent_signup` types module
- TypeScript SDK: 4 new static methods on `Inkbox` (`signup`, `verifySignup`, `resendSignupVerification`, `getSignupStatus`) + new `agent_signup` types with Raw/parse pattern
- CLI: new `inkbox signup` command group with `create`, `verify`, `resend-verification`, and `status` subcommands
- Documentation: Agent Signup sections added to all 3 SKILL files, all 4 SDK/CLI READMEs, and the top-level README

## Design decisions

- Signup methods are classmethods/static (not instance methods) because the agent doesn't have an API key until after `signup()` returns one — constructing a full `Inkbox` instance would require an API key that doesn't exist yet
- Post-signup methods (`verify`, `resend`, `status`) are also classmethods/static to keep the bootstrap flow lightweight — no need to spin up 6 HTTP transport pools just to submit a verification code
- `signup create` is the only CLI command that doesn't require `--api-key`

### DIFF
--- a/sdk/python/inkbox/agent_signup/types.py
+++ b/sdk/python/inkbox/agent_signup/types.py
@@ -1,0 +1,101 @@
+"""
+inkbox/agent_signup/types.py
+
+Dataclasses for the agent self-signup flow.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass
+class AgentSignupResponse:
+    """Response from ``POST /api/v1/agent-signup``."""
+
+    email_address: str
+    organization_id: str
+    api_key: str
+    agent_handle: str
+    claim_status: str
+    human_email: str
+    message: str
+
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> AgentSignupResponse:
+        return cls(
+            email_address=d["email_address"],
+            organization_id=d["organization_id"],
+            api_key=d["api_key"],
+            agent_handle=d["agent_handle"],
+            claim_status=d["claim_status"],
+            human_email=d["human_email"],
+            message=d["message"],
+        )
+
+
+@dataclass
+class AgentSignupVerifyResponse:
+    """Response from ``POST /api/v1/agent-signup/verify``."""
+
+    claim_status: str
+    organization_id: str
+    message: str
+
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> AgentSignupVerifyResponse:
+        return cls(
+            claim_status=d["claim_status"],
+            organization_id=d["organization_id"],
+            message=d["message"],
+        )
+
+
+@dataclass
+class AgentSignupResendResponse:
+    """Response from ``POST /api/v1/agent-signup/resend-verification``."""
+
+    message: str
+
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> AgentSignupResendResponse:
+        return cls(message=d["message"])
+
+
+@dataclass
+class SignupRestrictions:
+    """Behavioral restrictions applied to an agent based on claim status."""
+
+    max_sends_per_day: int
+    allowed_recipients: list[str]
+    can_receive: bool
+    can_create_mailboxes: bool
+
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> SignupRestrictions:
+        return cls(
+            max_sends_per_day=d["max_sends_per_day"],
+            allowed_recipients=d["allowed_recipients"],
+            can_receive=d["can_receive"],
+            can_create_mailboxes=d["can_create_mailboxes"],
+        )
+
+
+@dataclass
+class AgentSignupStatusResponse:
+    """Response from ``GET /api/v1/agent-signup/status``."""
+
+    claim_status: str
+    human_state: str
+    human_email: str
+    restrictions: SignupRestrictions
+
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> AgentSignupStatusResponse:
+        return cls(
+            claim_status=d["claim_status"],
+            human_state=d["human_state"],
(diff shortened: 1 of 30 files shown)


skills/inkbox-openclaw/SKILL.md
alex-w-99, line 1: Thoughts on creating `skills/agent-signup/SKILL.md`? Would consolidate the repeated agent signup instructions across `skills/inkbox-openclaw/SKILL.md`, `skills/inkbox-python/SKILL.md`, and `skills/inkbox-ts/SKILL.md`. Also useful because updates to that API would only require one update in one place.

Could reference it in those other skill files if necessary.

sdk/python/inkbox/client.py
```diff
@@ -249,3 +255,156 @@ def create_signing_key(self) -> SigningKey:
...
+            raise InkboxAPIError(status_code=resp.status_code, detail=str(detail))
+        return resp.json()
+
+    @classmethod
+    def signup(
+        cls,
+        human_email: str,
+        display_name: str,
+        *,
+        note_to_human: str | None = None,
```
alex-w-99: This is actually a required parameter. Will have to be updated here and everywhere else.

```suggestion
        note_to_human: str,
```

alex-w-99:
Can we please add Vitests+pytests for the agent signup flow (using mocks)?

sdk/python/inkbox/client.py
rayruizhiliao: good catch. fixed in commit https://github.com/inkbox-ai/inkbox/pull/30/commits/ec6a0bca8ac99aad6d526dcd40a0a6a0156b4368

sdk/python/inkbox/agent_signup/types.py
```diff
@@ -0,0 +1,101 @@
...
+
+@dataclass
+class AgentSignupResendResponse:
+    """Response from ``POST /api/v1/agent-signup/resend-verification``."""
+
+    message: str
+
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> AgentSignupResendResponse:
+        return cls(message=d["message"])
```
alex-w-99: `AgentSignupResendResponse` in both SDKs (TS+py) drops `claim_status` and `organization_id` from the response. The server returns the same data model (`AgentClaimResponse`) for both `/verify` and `/resend-verification`, but the verify SDK type captures all three fields while the resend type only keeps `message` attriute.

Should add `claim_status` and `organization_id` to both to match the actual server response.

skills/inkbox-openclaw/SKILL.md
rayruizhiliao: made the updates as suggested in commit https://github.com/inkbox-ai/inkbox/pull/30/changes/78b2613b68f601245b7af48a25d8c599abcf777b

rayruizhiliao:
> Can we please add Vitests+pytests for the agent signup flow (using mocks)?

added in commit https://github.com/inkbox-ai/inkbox/pull/30/commits/4f6584d211bf0f01ae72d6aff461731f04d5b767

sdk/python/inkbox/agent_signup/types.py
rayruizhiliao: updated in commits https://github.com/inkbox-ai/inkbox/pull/30/commits/2eb25c46d02e67e704fca8afdc8603fb67d72dc6 and https://github.com/inkbox-ai/inkbox/pull/30/commits/ba0ad85d8a292263d7c3252a3ae921a21db3f142

skills/agent-signup/SKILL.md
```diff
@@ -0,0 +1,198 @@
...
+### Verify
+
+```bash
+curl -X POST https://inkbox.ai/api/v1/agent-signup/verify \
+  -H "Content-Type: application/json" \
+  -H "X-Service-Token: ik_live_..." \
+  -d '{ "verification_code": "483921" }'
+```
+
+The verification code expires after 24 hours. Max 5 attempts before a resend is required.
```
alex-w-99:
```suggestion
The verification code expires after 48 hours. Max 5 attempts before a resend is required.
```

sdk/typescript/src/agent_signup/types.ts
```diff
@@ -0,0 +1,158 @@
...
+
+export function agentSignupRequestToWire(
+  req: AgentSignupRequest,
+): Record<string, string> {
+  const body: Record<string, string> = {
+    human_email: req.humanEmail,
+    display_name: req.displayName,
+  };
+  body["note_to_human"] = req.noteToHuman;
+  return body;
```
alex-w-99: Not that important but can be simplified now that `note_to_human` not optional.

```suggestion
  return {
    human_email: req.humanEmail,
    display_name: req.displayName,
    note_to_human: req.noteToHuman,
  };
```

skills/agent-signup/SKILL.md
rayruizhiliao: fixed in commit https://github.com/inkbox-ai/inkbox/pull/30/commits/cc4cb6d1388f9b6413eb94091a9393f33cd5d133

sdk/typescript/src/agent_signup/types.ts
rayruizhiliao: updated in commit https://github.com/inkbox-ai/inkbox/pull/30/commits/12f4c1541e76b30c3ae49fec90521c45e1f52ec2

alex-w-99 (approved):
LGTM

PR #31, opened 2026-04-06
Add SDK integration tests (Python, TypeScript, CLI)

dimavrem22:
## Summary
- Adds end-to-end integration tests exercising the Python SDK, TypeScript SDK, and CLI against live dev/beta environments
- New workflow (`sdk_integration.yml`) runs 6 jobs: 3 SDK types × 2 environments
- PR pushes test both development + beta; scheduled runs (every 2h) test beta only
- All base URLs are kept as secrets since this is a public repo
- Google Chat notifications on both scheduled success and failure
- Tests cover: whoami, identity CRUD, email send + delivery polling, message detail, threading, mark-read, identity pause/resume, signing key rotation

### DIFF
--- a/.github/workflows/sdk_integration.yml
+++ b/.github/workflows/sdk_integration.yml
@@ -58,6 +58,7 @@ jobs:
 
   python-sdk:
     name: Python SDK (${{ matrix.environment }}, ${{ matrix.source }})
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
     needs: select-targets
     runs-on: ubuntu-latest
     timeout-minutes: 25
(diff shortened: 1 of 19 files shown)


.github/workflows/sdk_integration.yml
alex-w-99, line 1: This workflow runs on every PR, but these jobs hard-require repo secrets. That's fine for same-repo branches like this one, but forked PRs won’t get those secrets, so they'll show failing/skipped-red integration checks even when the code is otherwise valid. 

I'd suggest gating the secret-backed jobs to same-repo PRs only, and relying on `workflow_dispatch` / local runs for forks when we want to validate them. That keeps contributor PRs from failing purely due to GitHub's secret model while still letting us run the integration suite manually before merge.

Something like:

```yaml
  if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
```

dimavrem22: Good call — added the fork gate to all three secret-dependent jobs (python-sdk, typescript-sdk, cli). Forked PRs will now skip these cleanly instead of failing on missing secrets.

dimavrem22: addressed

alex-w-99 (approved):

PR #76, opened 2026-06-17
`0.4.9`: Add Rust SDK (port of Python + TypeScript SDKs)

dimavrem22:
## Rust SDK

Adds a Rust SDK under `sdk/rust/` (crate `inkbox`), a faithful port of the existing Python (`inkbox`) and TypeScript (`@inkbox/sdk`) SDKs. Wire shapes — JSON field names, enum values, request bodies, query params, paths — match the other two SDKs exactly, since all three talk to the same server. The public surface is **blocking** (`reqwest::blocking`) to mirror the synchronous Python/TS APIs.

Versions move to **0.4.9** in lockstep across all three SDKs + CLI.

### What's included

- **Mail** — mailboxes, messages (send/forward/list/get/flags), threads, contact rules, domains.
- **Phone** — calls, numbers, texts (+ conversations), transcripts, contact rules, SMS opt-ins.
- **iMessage** — send/list/conversations/reactions/typing/media, contact rules.
- **Contacts / Notes** — CRUD, access grants, vCard import/export.
- **Identities** — `IdentitiesResource` + the `AgentIdentity` per-identity facade (channel-scoped `send_email`, `place_call`, `send_text`, `send_imessage`, credential/secret/TOTP helpers, …).
- **Vault** — Argon2id KDF + AES-256-GCM envelope encryption, TOTP (SHA-1/256/512), typed credential accessors. Parameters match Python/TS byte-for-byte, so secrets are cross-SDK interoperable.
- **API keys, webhooks** (receiver payload types + HMAC-SHA256 `verify_webhook`), **agent signup**, **whoami**.
- **Tunnels** — control plane (`list`/`get`/`update`/`sign_csr`) always available; the full data-plane runtime (h2/TLS + P-256 PKCS#10 CSR, HTTP / WS-upgrade / TCP-passthrough bridges) behind the optional `tunnels-runtime` feature.
- `publish.sh` mirroring the Python/TS/CLI publish scripts (dry-run by default, `--prod` to publish to crates.io).

### Verification

- `cargo check` and the full test suite pass with and without `--features tunnels-runtime` (rustls pinned to the ring backend — no aws-lc/cmake). **115 unit tests + doc-tests, all green.**
- Vault crypto and TOTP have `#[cfg(test)]` unit tests ported from the Python/TS suites, including a **cross-SDK `auth_hash` vector** that confirms Argon2id+SHA-256 output matches Python/TS byte-for-byte.
- **Live-validated end-to-end against production:**
  - **Vault** — initialize → unlock → encrypt+store a secret → fetch → decrypt round-trip, plus decrypting a secret created by another SDK (cross-SDK interop, both directions).
  - **Passthrough tunnel data plane** — real inbound HTTPS (GET/POST/concurrent) routed through the Rust agent's TLS termination to a local upstream, with clean TLS close.
  - **Real-time phone call** — provisioned a number, auto-accept → media WebSocket carried over the passthrough tunnel; an inbound call's bidirectional speech (greeting + transcript + reply) flowed through the Rust runtime.

The passthrough data plane is now production-complete; the earlier P-256 CSR and connection-orchestration work landed and was validated against the live edge.

### Note for reviewers

The branch also carries the merged Dependabot security bumps from #77 (esbuild / cryptography / idna / pytest / Pygments), which is the source of the TS/Python/CLI lock-file churn in the diff — unrelated to the Rust SDK. Happy to rebase those out onto `main` first if preferred.


cli/package.json
alex-w-99, line 1: `cli/src/index.ts:26` still reports `.version("0.4.8")`, so `inkbox --version` prints the wrong version despite `cli/package.json` saying `0.4.9`. Breaks the "all SDKs + CLI to 0.4.9 in lockstep" claim.

CHANGELOG.md
alex-w-99, line 1: `cli/src/commands/number.ts:201` still defaults `--type` to `"toll_free"`. Per the PR's own changelog the server now rejects `toll_free` (HTTP 422), so `inkbox number provision --handle X` (no `--type`) will fail outright.

sdk/python/uv.lock
alex-w-99, line 1: The `inkbox` self-version in `uv.lock` is `0.4.8` while `pyproject.toml` is `0.4.9`. `uv` will flag this on the next lock/sync.

alex-w-99:
Few additional thoughts/concerns:
- Looks like there aren't any Rust SDK related GitHub actions? I.e., linter or tests.
- Tunnel runtime reconnect triggers do not actually reconnect. In [runtime.rs](/home/ec2-user/inkbox/sdk/rust/src/tunnels/client/runtime.rs:333), the PING timeout branch just exits the spawned ping task. It does not close the h2 connection or notify `run_once`, which is waiting on `conn_closed` at [runtime.rs](/home/ec2-user/inkbox/sdk/rust/src/tunnels/client/runtime.rs:260). Same pattern for owner-token rejection: [runtime.rs](/home/ec2-user/inkbox/sdk/rust/src/tunnels/client/runtime.rs:422) returns from the intake task, but nothing forces the supervised connection down. Python explicitly calls `_force_reconnect_conn` for both cases. This can leave the Rust tunnel permanently stuck on a dead or invalid connection instead of reconnecting.
- Rust publish script does not load repo-root `.env` as documented. [publish.sh](/home/ec2-user/inkbox/sdk/rust/publish.sh:10) uses `$(dirname "$0")/../.env`, which resolves to `sdk/.env`, not repo-root `.env`. `RELEASING.md` says to put `CARGO_REGISTRY_TOKEN` in repo-root `.env`, so `cargo publish --prod` can fail unless the token is already in the environment.
- Medium/Low: published Rust README contradicts the PR/changelog validation claim. `CHANGELOG.md` says the passthrough tunnel data plane is live-validated, but [sdk/rust/README.md](/home/ec2-user/inkbox/sdk/rust/README.md:129) still says only 61 tests and [sdk/rust/README.md](/home/ec2-user/inkbox/sdk/rust/README.md:131) says the bridge pumps “have not been exercised against a live edge.” This is likely stale, but it will be published to crates.io.
- Root README still recommends retired toll-free provisioning. The PR updates SDK READMEs, but root [README.md](/home/ec2-user/inkbox/README.md:31) and [README.md](/home/ec2-user/inkbox/README.md:68) still use `type="toll_free"` / `{ type: "toll_free" }`, while the same PR says the server now rejects that with 422.

cli/package.json
dimavrem22: Fixed in 507f295 — `cli/src/index.ts` now reports `0.4.9`.

CHANGELOG.md
dimavrem22: Fixed in 9d967a5 — `number provision --type` now defaults to `local`.

sdk/python/uv.lock
dimavrem22: Fixed in 138fd83 — bumped the `inkbox` self-version in `uv.lock` to `0.4.9`.

dimavrem22:
All five addressed (one commit each):

- **No Rust CI** → 082575a — added a Rust job (build + test both feature sets; clippy/fmt advisory).
- **Reconnect never fires** → 82a68b0 — a per-connection `force_down` notify fired by the PING task and intake loops now breaks `run_once` so it reconnects (mirrors Python `_force_reconnect_conn`).
- **publish.sh wrong `.env`** → f05c19c — now loads repo-root `.env` (`../../.env`).
- **Stale Rust README** → 7a75254 — updated to 115 tests and live-validated passthrough.
- **Root README toll_free** → 16178a0 — examples now use the default `local`.

alex-w-99:
The Rust job builds both feature sets, but only runs `cargo test --features tunnels-runtime`; can we also run plain `cargo test` so CI covers the default/no-feature test path claimed in the PR?

dimavrem22:
Good catch — fixed in fe8b048; CI now runs both `cargo test` (the 56 default-feature vault/TOTP tests) and `cargo test --features tunnels-runtime`.

alex-w-99:
Looks like we've got to update the CI tests given the new identity limits. That's not related to this PR, however, so approving in spite of failing SDK integration tests.

alex-w-99 (approved):

PR #79, opened 2026-06-25
feat: add harness param to agent signup (SDK + CLI + skill)

dimavrem22:
Part of a cross-repo agent-signup `harness` change. This PR covers the public client surface: SDKs, CLI, and the self-signup skill.

## What
- **Optional `harness` request field** on agent self-signup — a free-form string naming the calling agent runtime (e.g. `claude-code`/`cc`, `codex`, `openclaw`, `hermes`, `cursor`, `windsurf`, etc.). Any value is accepted.
- **Signup response** now echoes `harness` (string|null) and adds `plugin_available` (bool).
- **Verify response** now returns `next_steps` (string).

## Where
- `sdk/python` — `Inkbox.signup()` gains a `harness` kwarg; `AgentSignupResponse` gains `harness`/`plugin_available`; verify response gains `next_steps`.
- `sdk/typescript` — `harness` on the request, `harness`/`pluginAvailable` on the response, `nextSteps` on verify; wire converters updated.
- `cli` — `signup create --harness <harness>`.
- `skills/inkbox-agent-self-signup` — documents the field and the new response fields, and instructs the agent: when verify returns and `plugin_available` is true, ask the human for permission before installing the matching plugin, then follow `next_steps` (which links the proper plugin repo).
- Integration tests round-trip `harness` in both SDKs.

The four published plugins referenced by the skill:
- https://github.com/inkbox-ai/claude-code-plugin
- https://github.com/inkbox-ai/codex-plugin
- https://github.com/inkbox-ai/openclaw-plugin
- https://github.com/inkbox-ai/hermes-agent-plugin

All new response fields are parsed defensively (default to null/false) so the SDKs stay compatible with servers that don't yet return them.

## Related PRs (agent-signup `harness`)
- inkbox-ai/inkbox#79 — SDK (py/ts) + CLI + self-signup skill (adds the `harness` param this builds on)
- inkbox-ai/claude-code-plugin#17
- inkbox-ai/codex-plugin#5
- inkbox-ai/openclaw-plugin#20
- inkbox-ai/hermes-agent-plugin#19

---
**Folds in #80** (`fix/whoami-iso-timestamps`): the Rust SDK now decodes whoami timestamps as ISO-8601 strings (tolerating a legacy epoch number) instead of `Option<f64>`, which failed to deserialize the server's string timestamps. Squashed here so the Rust SDK ships as one PR.

### DIFF
--- a/.github/workflows/sdk_integration.yml
+++ b/.github/workflows/sdk_integration.yml
@@ -87,7 +87,13 @@ jobs:
 
   python-sdk:
     name: Python SDK (${{ matrix.environment }}, ${{ matrix.source }})
-    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    # select-targets uses always(), so a skipped wait-for-unit-tests (every
+    # schedule/dispatch run) would otherwise propagate down and skip this job.
+    # always() + an explicit success check keeps the real gates intact.
+    if: |
+      always() &&
+      needs.select-targets.result == 'success' &&
+      (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
     needs: select-targets
     runs-on: ubuntu-latest
     timeout-minutes: 25
@@ -163,7 +169,13 @@ jobs:
 
   typescript-sdk:
     name: TypeScript SDK (${{ matrix.environment }}, ${{ matrix.source }})
-    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    # select-targets uses always(), so a skipped wait-for-unit-tests (every
+    # schedule/dispatch run) would otherwise propagate down and skip this job.
+    # always() + an explicit success check keeps the real gates intact.
+    if: |
+      always() &&
+      needs.select-targets.result == 'success' &&
+      (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
     needs: select-targets
     runs-on: ubuntu-latest
     timeout-minutes: 25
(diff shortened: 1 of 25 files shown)


dimavrem22:
codex: The SDK PR still has test expectation drift.

The TypeScript unit job is failing because `sdk/typescript/tests/agent_signup.test.ts` still expects the old parsed shapes; `parseAgentSignupResponse()` now returns `harness: null` and `pluginAvailable: false` for older/raw responses, and `parseAgentSignupVerifyResponse()` returns `nextSteps: null`. Please update those expected objects or intentionally preserve the old shape.

Also, the new Python and TypeScript integration assertions send `harness="cc"` but expect the response harness to be `cc`. The server PR normalizes `cc` to `claude-code` and documents the signup response as a normalized echo, so these should assert `claude-code` (and probably `plugin_available` / `pluginAvailable` true) or use canonical `claude-code` in the request.

dimavrem22:
Both addressed in b8f82cd:

1. **TS unit drift.** Updated `sdk/typescript/tests/agent_signup.test.ts` expectations to the new parsed shapes (`harness: null`, `pluginAvailable: false`, `nextSteps: null`); vitest is green.
2. **Normalization assertion.** The Python + TS integration tests still send `harness="cc"` (to exercise normalization) but now assert the response harness is `claude-code` and the plugin-available flag is truthy.

dimavrem22:
**Follow-up — version + Rust parity (`01897bc`):** This branch had not actually bumped the version — both it and `main` sat at `0.4.9`, which is already the **Rust SDK** release. Since versions move in lockstep, the harness feature is now **`0.4.10`** across `inkbox` (Python), `@inkbox/sdk` (TS), `@inkbox/cli`, and the Rust crate (+ changelogs, + `uv.lock` self-version). Also added the **Rust SDK** harness parity that was missed: `Inkbox::signup` now takes `harness`, `AgentSignupResponse` gained `harness` + `plugin_available`, and the verify response gained `next_steps` (all `#[serde(default)]`).

alex-w-99 (approved):

PR #80, opened 2026-06-25
fix(sdk/rust): decode whoami timestamps as ISO strings, not f64

dimavrem22:
The whoami response's `created_at` / `last_used_at` / `expires_at` were typed `Option<f64>` in the Rust SDK, but the API serializes these `datetime` fields as **ISO-8601 strings** (e.g. `"2026-06-25T23:05:46.406390Z"`) like every other timestamp it returns. serde rejected the string with `invalid type: string ..., expected f64`, so `whoami()` failed to decode for any real key.

Python/TS declare the same fields as `float`/`number` but don't enforce annotations at runtime, so they silently accept the string — Rust's serde is the only one that surfaced the long-standing type mismatch.

## Change
- Retype `created_at` / `last_used_at` / `expires_at` as `Option<String>` (the raw ISO timestamp), via a tolerant `deserialize_with` that accepts the ISO string and, defensively, a legacy epoch number.
- Fix the inaccurate `// Timestamps arrive as epoch floats` comment.
- Add regression tests: one decodes the exact ISO payload that failed, one proves numeric tolerance.

No version change. fmt / clippy (`-D warnings`, `tunnels-runtime`, all-targets) / `cargo test` (default + `tunnels-runtime`) all pass locally.

### DIFF
--- a/sdk/rust/src/whoami/types.rs
+++ b/sdk/rust/src/whoami/types.rs
@@ -30,13 +30,42 @@ pub struct WhoamiApiKeyResponse {
     pub label: Option<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub description: Option<String>,
-    // Timestamps arrive as epoch floats.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub created_at: Option<f64>,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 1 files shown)


dimavrem22:
Folded into #79 (cherry-picked commit f2cc5dca onto feat/agent-signup-harness) so the Rust SDK changes ship as a single PR. Closing in favor of #79.

PR #86, opened 2026-06-27
Add inbound email webhook example

Ayush7614:
## Summary

- Adds `examples/use-inkbox-webhook/` — inbound email webhook receiver using an in-process ASGI handler behind `inkbox.tunnels.connect()`
- Verifies webhook signatures with `verify_webhook()`, registers a `message.received` subscription, auto-replies once, and cleans up
- No ngrok or uvicorn required — tunnel exposes the handler at `https://{handle}.inkboxwire.com/hooks/mail`
- Updates the root README examples table

## Why

Polling (`04-inbox-monitor.sh`) exists but nothing demonstrated push-based inbound email handling. This is the production pattern for agents that react to mail as it arrives.

## Test plan

- [x] `ruff check` on `webhook_server.py` — clean
- [x] `pytest tests/test_webhook_subscriptions.py tests/test_signing_keys.py` — 43 passed
- [x] Live API run with console `INKBOX_API_KEY` — full end-to-end success (~12s):
  - Created identity `webhook-email-demo` (`webhook-email-demo@inkboxmail.com`)
  - Tunnel connected at `https://webhook-email-demo.inkboxwire.com`
  - Webhook subscription created for `message.received`
  - Probe email triggered inbound webhook
  - Signature verified via `verify_webhook()`
  - Auto-reply sent
  - Subscription and identity deleted on cleanup

## How to run

```bash
cd examples/use-inkbox-webhook
cp .env.example .env   # set INKBOX_API_KEY from inkbox.ai/console

cd ../../sdk/python
uv run --env-file ../../examples/use-inkbox-webhook/.env \
  python ../../examples/use-inkbox-webhook/webhook_server.py
```

Optional: set `INKBOX_WEBHOOK_SIGNING_KEY` to avoid calling `create_signing_key()` (which rotates the org signing key).

### DIFF
--- a/examples/use-inkbox-webhook/webhook_server.py
+++ b/examples/use-inkbox-webhook/webhook_server.py
@@ -5,9 +5,9 @@ Exposes an in-process ASGI handler at a public Inkbox tunnel URL, registers a
 ``message.received`` webhook subscription, sends a probe email, verifies the
 incoming webhook signature, auto-replies once, then cleans up.
 
-Requires INKBOX_API_KEY in the environment (see .env.example).
-Optional INKBOX_WEBHOOK_SIGNING_KEY — if unset, create_signing_key() is called
-(which rotates the org signing key).
+Requires INKBOX_API_KEY and INKBOX_WEBHOOK_SIGNING_KEY in the environment
+(see .env.example). Set INKBOX_ROTATE_SIGNING_KEY=1 only when you intend to
+rotate the org signing key via create_signing_key().
 """
 
 from __future__ import annotations
@@ -16,12 +16,12 @@ import asyncio
 import json
 import os
 import sys
+import uuid
 from typing import Any, cast
 
 from inkbox import Inkbox, MailWebhookPayload, verify_webhook
 from inkbox.exceptions import InkboxAPIError
 
-HANDLE = "webhook-email-demo"
 WEBHOOK_PATH = "/hooks/mail"
 PROBE_SUBJECT = "Webhook probe"
 WAIT_SECONDS = 90
@@ -33,6 +33,12 @@ def _require_env(name: str) -> str:
         print(f"ERROR: {name} is required.", file=sys.stderr)
         sys.exit(1)
     return value
+
+
+def _make_handle() -> str:
+    base = os.environ.get("INKBOX_AGENT_HANDLE", "webhook-demo").strip()
+    suffix = uuid.uuid4().hex[:8]
+    return f"{base}-{suffix}"
 
 
 async def _read_body(receive: Any) -> bytes:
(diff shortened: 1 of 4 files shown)


Ayush7614:
CC: @alex-w-99 @dimavrem22

examples/use-inkbox-webhook/webhook_server.py
```diff
@@ -0,0 +1,209 @@
...
+        {
+            "type": "http.response.start",
+            "status": status,
+            "headers": [(b"content-type", content_type.encode())],
+        },
+    )
+    await send({"type": "http.response.body", "body": body})
+
+
+def _ensure_identity(inkbox: Inkbox, handle: str) -> Any:
```
dimavrem22: codex: This deletes any existing `webhook-email-demo` identity before creating the demo identity. Identity deletion cascades to the mailbox, tunnel, scoped API keys, and phone number, so running this example can destroy a real org resource; please use a generated/overridable handle and only delete identities created by this run.

examples/use-inkbox-webhook/webhook_server.py
```diff
@@ -0,0 +1,209 @@
...
+        print(f"=> Removing existing identity: {handle}")
+        existing.delete()
+    except InkboxAPIError:
+        pass
+    identity = inkbox.create_identity(handle, display_name="Webhook Demo")
+    print(f"=> Created identity: {identity.agent_handle} ({identity.email_address})")
+    return identity
+
+
+def _resolve_signing_secret(inkbox: Inkbox) -> str:
```
dimavrem22: codex: Defaulting to `create_signing_key()` rotates the org-level webhook signing key when `INKBOX_WEBHOOK_SIGNING_KEY` is absent. That can immediately break existing webhook receivers still verifying with the old key; please require the signing key or gate rotation behind an explicit flag.

examples/use-inkbox-webhook/webhook_server.py
```diff
@@ -0,0 +1,209 @@
...
+                    subject=f"Re: {PROBE_SUBJECT}",
+                    body_text="Got your webhook — auto-reply from the Inkbox webhook example.",
+                    in_reply_to_message_id=message_id,
+                )
+                print(f"=> Auto-replied to {from_address}")
+                webhook_received.set()
+
+            await _send_response(send, 200, b"ok")
+
+        print("=> Connecting tunnel (in-process ASGI handler)")
```
dimavrem22: codex: Cleanup only starts after subscription creation succeeds, so failures during tunnel connect/startup or subscription creation leak the newly created identity and tunnel. Please wrap the whole setup after identity creation in `try/finally` so failed runs also clean up.

examples/use-inkbox-webhook/webhook_server.py
```diff
@@ -0,0 +1,209 @@
...
+
+            if (
+                event_type == "message.received"
+                and direction == "inbound"
+                and subject == PROBE_SUBJECT
+                and not handled_probe
+            ):
+                handled_probe = True
+                from_address = message.get("from_address") or identity.email_address
+                message_id = message.get("id")
```
dimavrem22: codex: This uses the webhook row `id` as `in_reply_to_message_id`, but the webhook schema documents `message_id` as the RFC Message-ID and `id` as the stored row id. Use `message["message_id"]` here, or call `reply_all_email(message["id"], ...)`, otherwise the auto-reply may not thread correctly.

Ayush7614:
@dimavrem22 thanks for the review !

Addressed all four review comments in 10e58d2:

1. **Handle** — unique suffix via `INKBOX_AGENT_HANDLE` base (default `webhook-demo-{uuid}`); no longer deletes pre-existing identities
2. **Signing key** — `INKBOX_WEBHOOK_SIGNING_KEY` is required; rotation only when `INKBOX_ROTATE_SIGNING_KEY=1`
3. **Cleanup** — `try/finally` wraps the full run after identity creation so tunnel/subscription failures still delete the demo identity
4. **Threading** — auto-reply uses `reply_all_email(stored_message_id, ...)` with the webhook row `id`

README and `.env.example` updated to match.

Ayush7614:
Follow-up in `b56967d` — identity-scoped signing keys + live re-test:

**Signing key (review 2, updated for API change)**
Org-level `POST /signing-keys` is now deprecated (409 on live API). The example instead:
- Uses `INKBOX_WEBHOOK_SIGNING_KEY` from env when set
- Otherwise creates a **per-identity** key via `POST /identities/{handle}/signing-key` after provisioning the demo identity (shown once)
- `INKBOX_ROTATE_SIGNING_KEY=1` still gates explicit rotation

Only `INKBOX_API_KEY` is required in `.env` now; README and `.env.example` updated.

**Auto-reply (review 4, edge case)**
Live run verified webhook delivery + signature verification end-to-end. Auto-reply via `reply_all_email` is skipped for the self-sent probe email (`HTTP 422: No reply-all recipients`) — expected because the probe is sent from the same mailbox to itself. Handler catches this and still marks the webhook as verified; fallback uses `send_email` with `in_reply_to_message_id=message["message_id"]` when applicable.

**Live test result**
- Identity + identity signing key created
- Tunnel connected, subscription created
- Probe email → webhook received, signature verified
- Cleanup (subscription + identity) succeeded, exit 0

dimavrem22 (approved):
Review comments addressed; verified compatible with current main. Approving to merge.

PR #87, opened 2026-06-27
Webhooks: per-event id + delivery log & replay

rayruizhiliao:
## Webhooks: per-event `id` + delivery log & replay

Two webhook changes to the SDKs (Python / TypeScript / Rust) and CLI, tracking the `servers#webhooks` backend. Shipped lockstep as **0.4.14**.

### 1. Stable per-event `id` (`evt_…`)

Surfaces the new top-level `id` on the mail / text / iMessage webhook payload types so consumers can dedupe on it. Use `id` as the **idempotency key** — it's stable across retries and replays — whereas `X-Inkbox-Request-ID` is per-delivery.

- **Python:** `id: str` on the `MailWebhookPayload` / `TextWebhookPayload` / `IMessageWebhookPayload` TypedDicts.
- **TypeScript:** `id: string` on the corresponding interfaces.
- **Rust:** `id: String` on the corresponding structs, marked `#[serde(default)]` so a missing field deserializes to `""` instead of hard-failing.
- Incoming-call payload types are unchanged (flat, already carry a call `id`).

> **Why `#[serde(default)]` on the Rust field:** unlike the Python `TypedDict` / TS `interface` (compile-time only — `json.loads` / `JSON.parse` don't enforce them), serde's derived `Deserialize` treats a bare `id: String` as **required**, so a consumer would hard-fail parsing any payload without it (an older or mixed-deployment server, or a recorded/mocked fixture predating the field). The `default` keeps the public type `String` (no source break, parity with Python `str` / TS `string`) while tolerating absence.

### 2. Delivery log + manual replay (new)

Every outbound webhook attempt is logged (signed request body, endpoint response or transport error, timing, replay flag). New resource to inspect the log and replay a missed delivery to its subscription's current URL.

- **SDKs:** `inkbox.webhooks.deliveries.list(...)` / `.replay(deliveryId)` (Python `inkbox.webhooks.deliveries`, Rust `inkbox.webhooks().deliveries()`), returning a new `WebhookDelivery` type. `list` filters on subscription / phone number / event type / 2xx success, with `limit` + `offset` paging.
- **CLI:** `inkbox webhook delivery list` (`--subscription-id` / `--phone-number-id` / `--event-type` / `--success` / `--failed` / `--limit` / `--offset`) and `inkbox webhook delivery replay <delivery-id>`.
- **Replay semantics:** reuses the original envelope `event_id`, so a compliant endpoint dedupes a replay it already processed — it recovers a *miss* rather than forcing reprocessing. Incoming-call deliveries are logged but **not replayable**.

### Versioning

- Python / TypeScript / Rust SDKs + CLI bumped to **0.4.14** in lockstep.
- Fixed the CLI's `@inkbox/sdk` dependency pin (was `^0.4.12`, now `^0.4.14`) and brought the stale Rust `Cargo.lock` in line.

### Backward compatibility

**Additive.** Existing code ignores the new `id` field until upgraded; the server already sends it. The new `deliveries` resource is purely additive. With the Rust `#[serde(default)]` fix above, payload deserialization no longer regresses against older or mixed-deployment servers in any of the three SDKs.

### Notes for reviewers

- **Docs** (delivery-log API reference + guide updates) land in the separate **website** repo PR.
- **Verification:** TypeScript SDK builds clean, 759 tests pass; CLI builds + tests pass; Python validated via an isolated round-trip test. The **Rust SDK was not compiled in the authoring environment** (no `cargo` available) — please run `cargo test` before un-drafting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
 Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
 (Python), and `@inkbox/cli`.
 
+## 0.4.14 — Webhook delivery log
+
+### Added
+
+- **Webhook delivery log + manual replay** in the TypeScript, Python, and Rust SDKs, plus the CLI. Every outbound webhook attempt is logged with its signed request body, the endpoint's response (or transport error), timing, and a replay flag.
+  - SDKs: `inkbox.webhooks.deliveries.list(...)` / `.replay(deliveryId)` (Python `inkbox.webhooks.deliveries`, Rust `inkbox.webhooks().deliveries()`), returning the new `WebhookDelivery` type. `list` filters on subscription, phone number, event type, and 2xx success, with `limit` / `offset` paging.
+  - CLI: `inkbox webhook delivery list` (with `--subscription-id` / `--phone-number-id` / `--event-type` / `--success` / `--failed` / `--limit` / `--offset`) and `inkbox webhook delivery replay <delivery-id>`.
+  - Replay reuses the original envelope `event_id`, so a compliant endpoint dedupes a replay it already processed — it recovers a miss rather than forcing reprocessing. Incoming-call deliveries are logged but not replayable.
+
+### Fixed
+
+- **Rust webhook payloads tolerate a missing `id`.** The per-event `id` added to `MailWebhookPayload` / `TextWebhookPayload` / `IMessageWebhookPayload` in 0.4.13 was a required serde field, so a 0.4.13 Rust consumer hard-failed deserializing any payload without it (an older or mixed-deployment server, or a recorded/mocked payload predating the field). The field is now `#[serde(default)]` — an absent `id` parses to `""`. The TypeScript and Python types were already runtime-tolerant (compile-time-only `interface` / `TypedDict`).
+
 ## 0.4.13 — Webhook event id
 
 ### Added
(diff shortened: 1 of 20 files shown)


alex-w-99:
The "additive / backward compatible" claim doesn't hold for the Rust SDK. For TypeScript (`interface`) and Python (`TypedDict`), the types are compile-time only — `JSON.parse` / `json.loads` don't enforce them, so a payload without `id` still parses and existing code is unaffected at runtime. But in `sdk/rust/src/webhooks/types.rs`, `pub id: String` on `MailWebhookPayload` / `TextWebhookPayload` / `IMessageWebhookPayload` is a **required serde field with no `#[serde(default)]`** (the existing structs use no `Option<>` fields), so deserialization **hard-fails if `id` is absent**. That breaks a 0.4.13 Rust consumer pointed at an older or mixed-deployment server, and breaks parsing of any recorded/mocked payload that predates the field. Can you please make the Rust field tolerant (`pub id: Option<String>` or `#[serde(default)]`)?

rayruizhiliao:
> The "additive / backward compatible" claim doesn't hold for the Rust SDK. For TypeScript (`interface`) and Python (`TypedDict`), the types are compile-time only — `JSON.parse` / `json.loads` don't enforce them, so a payload without `id` still parses and existing code is unaffected at runtime. But in `sdk/rust/src/webhooks/types.rs`, `pub id: String` on `MailWebhookPayload` / `TextWebhookPayload` / `IMessageWebhookPayload` is a **required serde field with no `#[serde(default)]`** (the existing structs use no `Option<>` fields), so deserialization **hard-fails if `id` is absent**. That breaks a 0.4.13 Rust consumer pointed at an older or mixed-deployment server, and breaks parsing of any recorded/mocked payload that predates the field. Can you please make the Rust field tolerant (`pub id: Option<String>` or `#[serde(default)]`)?

fixed and updated other pieces in commit https://github.com/inkbox-ai/inkbox/pull/87/commits/bda4b08feffa45b1f76d51fcd94558ab6abbf064

alex-w-99 (approved):

PR #90, opened 2026-07-01
`0.4.15`: feat(phone)! agent-identity-centered phone SDK (0.4.15) — drop number-scoped call/transcript paths

dimavrem22:
## What
Re-centers the phone module of all three SDKs (Python, TypeScript, Rust) on the **agent identity**, to match the identity-scoped phone API. Ships as a **`0.4.15`** release.

The deprecated **number-scoped** call/transcript surface is **fully removed** (no back-compat shims). The underlying API remains backwards-compatible for existing callers of the old endpoints; the SDK simply targets the identity surface going forward.

## Surface (the 4 things, identical across languages)
- **Calls are identity-scoped:** `calls.list(agent_identity_id=…)`, `calls.get(call_id)`, `calls.transcripts(call_id)`. The standalone transcripts resource is removed and folded into `calls`.
- **place-call:** gains `origination` (`dedicated_number` | `shared_imessage_number`) + `agent_identity_id`; `from_number` is now conditional (dedicated only). Agent-scoped keys resolve their own identity, so `agent_identity_id` is optional for them.
- **incoming-call-action:** new resource (`get`/`set`) for the identity's inbound-call config, plus convenience delegators on `AgentIdentity` in all three languages.
- **Response types:** `PhoneCall` gains `origin`; `local_phone_number` is now **nullable** (null for shared-line calls — the underlying shared line is never surfaced). Rust uses `Option<String>` with a serde default.

## Removed
The number-scoped `GET /numbers/{id}/calls`, `/calls/{id}`, `/calls/{id}/transcripts` bindings and the standalone transcripts resource, across Python/TS/Rust.

## Versioning
`0.4.14 → 0.4.15` in `pyproject.toml`, `package.json`, `Cargo.toml`.

## Verification
- Unit suites green: **Python 739**, **TypeScript 796**, **Rust 62**.
- Live acceptance of all four surfaces (list/get/transcripts, incoming-call-action get/set incl. 422 validation, place-call dedicated **and** shared→409) run against the development API in every language.


sdk/rust/Cargo.toml
```diff
@@ -8,7 +8,7 @@
 
 [package]
 name = "inkbox"
-version = "0.4.14"
+version = "1.0.0"
```
alex-w-99: I feel like this should just be `0.4.15`. Why the big jump?

alex-w-99:
Should pin SDK version in CLI's `package.json` file to this new release version.

sdk/rust/Cargo.toml
dimavrem22: Update: went with 0.4.15 per Dima — applied across all three SDKs, the CLI version, its `@inkbox/sdk` pin (`^0.4.15`), and the lockfiles in cb9eca8. PR title/description updated to match. For the record since it informs the release notes: consumers on `^0.4` ranges will pick this version up automatically, and the removed number-scoped methods mean that upgrade is source-breaking — worth a loud changelog entry when publishing.

alex-w-99:
Can you please add an entry for 0.4.15 to `CHANGELOG.md`?

dimavrem22:
@alex-w-99 Done — 0.4.15 entry added to the root CHANGELOG.md in 9332553 (Added / Changed / Removed, with the source-breaking note for `^0.4` upgraders). The per-package changelogs have been stale since 0.4.12, so I followed the 0.4.13/0.4.14 precedent of the root being canonical.

alex-w-99 (approved):

PR #92, opened 2026-07-04
SDK/CLI: configurable webhook context + open tracking (0.4.16)

alex-w-99:
SDK + CLI support for the configurable-webhooks server features:

- Per-class webhook `context_config` on subscription create/update, and typed `data.context` payload shapes for receivers.
- Email open tracking: `track_opens` on send/forward plus `first_opened_at`/`open_count` on messages.
- Read-on-fetch docs (API-key GET of an inbound message marks it read).
- Integration harness: per-run provisional-org cleanup.

Bumps `inkbox`, `@inkbox/sdk`, and `@inkbox/cli` to 0.4.16. Unit suites green (Python 798, TS 843, CLI 12).


sdk/python/inkbox/webhooks.py
```diff
@@ -119,6 +119,99 @@ class WebhookAgentIdentity(TypedDict):
...
+    omitted_ms: int
+
+
+class WebhookContextCallItemWire(TypedDict):
+    """One calls-class item: metadata plus its (possibly abridged) transcript."""
+    call_id: str
+    abridged: bool
+    transcript: list[WebhookTranscriptEntryWire]
+    direction: NotRequired[str]
+    counterparty: NotRequired[str]
```
dimavrem22: 🔴 **CRITICAL — call-context field names don't match the webhook wire format.** The delivered `data.context.calls.items[]` uses `remote_number` (the far-end number) and `duration` (seconds); this type declares `counterparty` and `duration_s`. Both are `NotRequired`, so `JSON.parse`/dict access still succeeds — but these two keys **never populate from a real payload**, so an agent reading call context can't see who the call was with or how long it ran.

Fix: `remote_number: NotRequired[str]` → in fact make it `remote_number: str | None` (see the present-as-null note below), and `duration: int | None` (the value is already whole seconds, so only the name is stale). Update the test fixtures alongside.

sdk/python/inkbox/webhooks.py
```diff
@@ -119,6 +119,99 @@ class WebhookAgentIdentity(TypedDict):
...
+
+
+class WebhookContextMailItemWire(TypedDict):
+    """Slim mail context item: metadata + snippet only; bodies are omitted."""
+    id: str
+    direction: str
+    from_address: str
+    to_addresses: list[str]
+    created_at: str
+    subject: NotRequired[str]
```
dimavrem22: 🟠 **HIGH — item fields are present-as-`null` on the wire, not absent.** The API emits item-level fields with the key **always present** (value `null` when unset), not omitted. This applies to `MailItem.subject`/`snippet`, `TextItem.sender`/`status`/`media`, and `CallItem.direction`/`remote_number`/`duration`/`started_at`.

Typing them `NotRequired[str]` says “may be absent; a `str` when present” — but the wire sends `str | None`, always present. They should be `subject: str \| None` (required key), etc. Note the **Block-level** optionals (`mode`/`requested`/`hours`/`skipped`) *are* pruned by the API, so `NotRequired` is correct there — just not for item fields. The class docstring's “optional keys are absent, not null” holds for the Block but is wrong for items.

sdk/python/inkbox/webhooks.py
```diff
@@ -431,6 +526,7 @@ class IMessageWebhookData(TypedDict):
     reaction: IMessageWebhookReaction | None
     contacts: list[WebhookContact]
     agent_identities: list[WebhookAgentIdentity]
+    context: NotRequired[WebhookContextWire]
```
dimavrem22, line 560: 🔵 **LOW — `context` is attached to the shared iMessage data type used by every `imessage.*` event**, but the API only delivers a context block on `imessage.received` — never `imessage.reaction_received` or delivery-status events. It's `NotRequired`, so this is tolerant (reaction payloads simply omit it), but the type lets `context` appear where it never will. Minor — a doc note or a narrower type would encode the exclusion.

sdk/python/tests/test_webhook_context_types.py
```diff
@@ -0,0 +1,159 @@
...
+    payload = cast(
+        MailWebhookPayload, json.loads(json.dumps(_message_received_with_context()))
+    )
+    ctx: WebhookContextWire = payload["data"]["context"]
+    assert set(ctx.keys()) == {"email", "texts", "calls"}
+    assert ctx["email"]["mode"] == "count"
+    assert ctx["email"]["requested"] == 2
+    assert ctx["email"]["items"][0]["from_address"] == "a@b.c"
+    # An item without subject/snippet simply omits the keys.
+    assert "subject" not in ctx["email"]["items"][0]
```
dimavrem22: This asserts `"subject" not in items[0]`, but the API always includes `subject` (as `null` when unset), so this bakes in a shape the server never emits and would fail against a real payload. Change to `assert ctx["email"]["items"][0]["subject"] is None` once the type is `subject: str \| None`. (Same applies to any other “key not in item” assertions for item fields.)

sdk/typescript/src/webhooks/types.ts
```diff
@@ -81,6 +81,96 @@ export interface WebhookAgentIdentity {
...
+  omitted_ms?: number;
+}
+
+/** One calls-class item: metadata plus its (possibly abridged) transcript. */
+export interface WebhookContextCallItem {
+  call_id: string;
+  abridged: boolean;
+  transcript: WebhookTranscriptEntry[];
+  direction?: string;
+  counterparty?: string;
```
dimavrem22: 🔴 **CRITICAL — stale call-context field names.** Same issue as the Python type: the wire sends `remote_number` and `duration` (seconds); this declares `counterparty?` and `duration_s?`. This module is a raw snake_case wire layer with no transform, so the type names *are* the wire names — and these two never match, so `item.remote_number` / `item.duration` come back `undefined`. Fix: `remote_number: string \| null`, `duration: number \| null`. Update the fixture at `tests/webhooks-types.test.ts:196-197` too.

sdk/typescript/src/webhooks/types.ts
```diff
@@ -81,6 +81,96 @@ export interface WebhookAgentIdentity {
...
+export interface WebhookContextTextItem {
+  id: string;
+  channel: WebhookContextTextChannelWire;
+  direction: string;
+  text: string;
+  text_truncated: boolean;
+  created_at: string;
+  sender?: string;
+  status?: string;
+  media?: Record<string, unknown>;
```
dimavrem22: 🟠 **HIGH — `media` shape is wrong.** The API sends `media: { count: number } \| null` (metadata count only — never URLs or attachment detail). `Record<string, unknown>` is both too loose and omits that the sole real key is a numeric `count`. Worse, the test fixture at `tests/webhooks-types.test.ts:184` uses `media: { content_type: "image/png", size: 1234 }`, an attachment shape the API never emits for context items. Fix the type to `media?: { count: number } \| null` and the fixture to `{ count: 1 }`.

sdk/typescript/src/webhooks/types.ts
```diff
@@ -81,6 +81,96 @@ export interface WebhookAgentIdentity {
...
+export type WebhookContextTextChannelWire = "sms" | "imessage";
+
+/** Slim mail context item: metadata + snippet only; bodies are omitted. */
+export interface WebhookContextMailItem {
+  id: string;
+  direction: string;
+  from_address: string;
+  to_addresses: string[];
+  created_at: string;
+  subject?: string;
```
dimavrem22: 🟠 **HIGH — item fields are present-as-`null` on the wire (runtime footgun).** Like the Python type, the API sends these item fields present-with-`null`, not absent — `MailItem.subject`/`snippet`, `TextItem.sender`/`status`, `CallItem.direction`/`remote_number`/`duration`/`started_at`. `subject?: string` is `string | undefined` and **can't hold `null`**, so a consumer that narrows `if (item.subject !== undefined) item.subject.toUpperCase()` crashes on a real `null` (`null !== undefined` is `true`). Fix: `subject: string \| null` (required key), etc. (Block-level `mode`/`requested`/`hours`/`skipped` stay optional — those are pruned by the API.)

sdk/typescript/tests/webhooks-types.test.ts
```diff
@@ -147,6 +149,160 @@ describe("MailWebhookPayload", () => {
...
+      ],
+      truncated: true,
+    },
+    calls: {
+      scope: "contact",
+      items: [
+        {
+          call_id: "call_1",
+          direction: "inbound",
+          counterparty: "+15551234567",
```
dimavrem22: This fixture sets `counterparty`/`duration_s`, locking in the stale call-item shape (it only passes because the type above is wrong). Update to `remote_number`/`duration` alongside the type fix. See also line 184's `media` fixture, which uses an attachment shape (`content_type`/`size`) the API never emits for context items — should be `{ count: 1 }`.

sdk/python/inkbox/mail/types.py
```diff
@@ -267,6 +267,11 @@ class Message:
     is_starred: bool
     has_attachments: bool
     created_at: datetime
+    # Open tracking (outbound sends with track_opens). ``first_opened_at`` is
+    # the signal; ``open_count`` is an upper bound (image proxies prefetch
```
dimavrem22: 🔵 **LOW — doc drift on `open_count`.** It's described as an “upper bound,” but the API now characterizes it as *approximate / biased both ways*: image-proxy prefetch inflates it, while the per-window debounce collapses multiple opens in a window to a single increment — so it can also read **below** the true open count, i.e. not strictly an upper bound. Worth matching that wording here (and at `resources/messages.py:127` and `README.md:232`). `first_opened_at` remains the reliable “was it ever opened” signal.

dimavrem22 (commented):
## SDK/CLI compliance audit vs the finalized webhook + mail API contract

Deep-dived Python + TypeScript + CLI (and Rust) and independently re-verified every finding against the **finalized** wire contract. **Not yet 100% compliant** — one CRITICAL (call-context field names, in both Python and TS) plus a couple of typing bugs. Inline comments below, most-severe first.

Two cross-cutting items that don't map to a single changed line:

**🟡 MEDIUM — Rust SDK has none of this feature.** `sdk/rust` ships no `context_config`, no `data.context` types, no `track_opens`, and no `first_opened_at`/`open_count`, and stays at `0.4.15`. There's no `#[serde(deny_unknown_fields)]` anywhere, so existing Rust receivers degrade gracefully (unknown fields ignored, nothing breaks) — but a Rust sender can't opt into open tracking and a Rust receiver can't read conversation context or open counts. Fine as a tracked follow-up; when it's added it should use `remote_number`/`duration` from day one (Rust already uses `remote_number` elsewhere).

**🔵 LOW — CLI docs/display.** `cli/README.md` omits the `forward` and `reply-all` commands entirely, so `email forward --track-opens` ships undocumented. And `openCount`/`firstOpenedAt` are shown in `email list`/`get` but not in `search`/`unread`/`thread` (the field names used are correct where shown — purely a display-completeness nit).

**Verified fully compliant (no action):** `context_config` validation (count 1–50 / window hours 1–168, discriminated union on `mode`, extra-key + unknown-class rejection, omit/null/object tri-state update); `track_opens` on both send and forward (send guards the no-HTML 422 trap, forward correctly defers to the server); `first_opened_at`/`open_count` names/types/defaults + null→0 coercion; read-on-fetch docs (API-key + inbound only, no overpromise); versions all `0.4.16` (+0.0.1) with root + per-package changelog entries.

sdk/python/inkbox/webhooks.py
alex-w-99: Fixed → item fields now required `str | None` (present-as-null) in 898e242.

sdk/python/inkbox/webhooks.py
alex-w-99: Fixed → `remote_number`/`duration` in 898e242.

sdk/python/inkbox/mail/types.py
alex-w-99: Reworded (approximate/biased both ways) across all spots in 898e242.

sdk/typescript/tests/webhooks-types.test.ts
alex-w-99: Updated (`remote_number`/`duration`, `media: { count: 1 }`) in 898e242.

sdk/typescript/src/webhooks/types.ts
alex-w-99: Fixed → item fields now required `string | null` in 898e242.

sdk/typescript/src/webhooks/types.ts
alex-w-99: Fixed → `{ count: number } | null` + fixture `{ count: 1 }` in 898e242.

sdk/python/tests/test_webhook_context_types.py
alex-w-99: Updated to `is None` + present-as-null fixture in 898e242.

sdk/typescript/src/webhooks/types.ts
alex-w-99: Fixed → `remote_number`/`duration` in 898e242.

sdk/python/inkbox/webhooks.py
alex-w-99: Doc note added on all three data types in 898e242.

alex-w-99:
Cross-cutting: CLI docs (`forward`/`reply-all`) + `openCount`/`firstOpenedAt` in `search`/`unread`/`thread` fixed in 898e242. Rust left as the follow-up you flagged.

dimavrem22 (approved):
Approving — all review comments addressed and re-verified against the finalized webhook + mail API contract, and Rust parity now landed.

**Review findings (Python/TS/CLI), all fixed & verified in `898e242`:**
- Call-context field names corrected to `remote_number` / `duration` (were `counterparty` / `duration_s`) in Python + TS + tests.
- Present-as-null item fields now typed as required-nullable (Python `str | None`, TS `string | null` with "narrow with === null"); fixtures/asserts updated to present-with-null.
- `media` typed `{ count }` (Python/TS) + test fixtures fixed.
- `open_count` doc reworded from "upper bound" to approximate/biased-both-ways.
- Reaction over-attach documented; CLI README (forward/reply-all + --track-opens) and open-field display added.

**Rust parity (`11bb00d`) verified compliant:** `WebhookContextCallItem` uses `remote_number`/`duration` (Option); scope/skip/channel/mode enums via `#[serde(rename_all = "snake_case")]`; `context_config` on subscriptions; `track_opens` on send + forward (HTML-required, 422); `first_opened_at`/`open_count` on the message response; Cargo bumped 0.4.15 → 0.4.16. Rust tests green.

**One optional, non-blocking nit:** Rust models texts `media` as `Option<serde_json::Value>` rather than a typed `{ count }` (as Python/TS do). It deserializes null/`{count}` correctly, so it's fine — just looser than the other two if you want exact cross-language parity.

CI green across Python/TS/CLI/Rust. LGTM.

PR #94, opened 2026-07-04
`0.4.17`: Inline images, CLI attachments, and mark-unread

alex-w-99:
Adds inline-image support and closes a few SDK/CLI parity gaps. Bumps all four packages to `0.4.17`.

- `messages.send`/`reply_all` (and identity `send_email`/`reply_all_email`) attachments accept `content_id` (TS `contentId`, Rust `Attachment.content_id`) to render an image inline in the HTML body via `cid:<content_id>`. Requires `body_html`, an `image/*` content type, and a unique id per send. Not supported on forwards.
- Adds `mark_emails_unread` (TS `markEmailsUnread`) to the identity across the TypeScript, Python, and Rust SDKs, the batch counterpart to `mark_emails_read`.
- CLI gains `--attach` on `email send`/`reply-all`/`forward` and `--inline-image <cid=path>` on `send`/`reply-all`, plus `email mark-unread` and `email download-attachment` commands.
- Tests, READMEs, skills, and the changelog updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
 Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
 (Python), `@inkbox/cli`, and `inkbox` (Rust, crates.io).
 
+## 0.4.17 — Inline images, CLI attachments, and mark-unread
+
+### Added
+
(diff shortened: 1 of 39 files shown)


dimavrem22 (approved):
Approving — 0.4.17 SDK/CLI surface is clean and consistent across Py/TS/Rust + CLI. Version-drift guards added, Rust source-break documented in the changelog, and local inline-image validation tightened. 🚀

PR #95, opened 2026-07-06
`0.4.18`: feat(tunnels); stop and don't reconnect when another client takes over

alex-w-99:
When another client connects to the same tunnel, the earlier client now stops and does not reconnect. Previously it would redial, and two clients on one tunnel could bounce back and forth. Run one client per tunnel; use separate identities for redundancy.

- New `"superseded"` runtime status plus a distinct terminal outcome in all three tunnel clients (Python, TypeScript, Rust). TypeScript exports `TunnelSupersededError`; Python stops out of `serve()` / `wait()`; Rust returns a terminal error from `serve_forever`.
- A normal server redeploy still reconnects seamlessly.
- Tests added for each runtime; SDKs + CLI bumped to `0.4.18`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@ Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
 
 ### Changed
 
-- **A tunnel keeps a single live client; a newer client takes over.** When another client connects to the same tunnel, the earlier client now stops and does **not** reconnect — previously it would redial, and two clients on one tunnel could bounce back and forth. Run one client per tunnel; for redundancy, use separate identities. The runtime reports a new `"superseded"` status through the status callback (`onStatus` / `on_status`) and ends with a distinct terminal outcome instead of retrying: TypeScript throws `TunnelSupersededError`, Python stops out of `serve()` / `wait()`, and Rust returns a terminal `Err` from `serve_forever`. A normal server redeploy still reconnects seamlessly, unchanged.
+- **A tunnel keeps a single live client; a newer client takes over.** When another client connects to the same tunnel, the earlier client now stops and does **not** reconnect; previously it would redial, and two clients on one tunnel could bounce back and forth. Run one client per tunnel; for redundancy, use separate identities. The runtime reports a new `"superseded"` status through the status callback (`onStatus` / `on_status`) and ends with a distinct terminal outcome instead of retrying: TypeScript throws `TunnelSupersededError`, Python raises `TunnelSupersededError` out of `serve()` / `wait()`, and Rust returns a terminal `Err` from `serve_forever`. A normal server redeploy still reconnects seamlessly, unchanged.
 
 ### Added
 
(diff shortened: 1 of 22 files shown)


dimavrem22:
codex: @alex-w-99 the latest SDK commit fixes the over-broad handoff predicate, but one takeover retry path still looks open.

Python now treats every connection as terminal except one explicitly tracked in `self._draining` (`sdk/python/inkbox/tunnels/client/_runtime.py:760`), and TypeScript mirrors that with `!this.draining.has(conn)` (`sdk/typescript/src/tunnels/client/_runtime.ts:565`). That fixes the non-active replacement being misclassified as self-handoff.

The remaining gap is when the replacement hello is interrupted by a superseded GOAWAY but the hello awaiter sees only a plain reset/status-0 error. In Python, the GOAWAY handler sets `_superseded` (`_runtime.py:1280`), `_await_response()` can return status `0` on the synthetic reset (`_runtime.py:971`), and `_send_hello()` then raises a plain transient error. `_make_replacement_connection()` catches that plain exception and retries unless the exception itself is `_TunnelAuthError` or `_TunnelSupersededError` (`_runtime.py:697`), unlike the outer cold-connect loop which checks `self._superseded` before reconnecting (`_runtime.py:500`). TypeScript has the same shape: `serveForever()` checks `this.superseded` (`_runtime.ts:370`), but `makeReplacementConnection()` catches plain errors internally and retries without checking the flag (`_runtime.ts:691`).

I verified this against the latest head with a focused Python repro: monkeypatch `_send_hello()` to set `runtime._superseded = True` and raise a plain `RuntimeError` on the first replacement attempt; `_make_replacement_connection()` made a second hello attempt (`attempts=2 superseded=True`). Once a terminal takeover signal has been observed, the handoff helper needs to stop immediately regardless of whether the current await surfaced a typed takeover exception or a generic reset.

Suggested fix: in both Python and TypeScript, check the terminal flag in `makeReplacementConnection()` before retrying any plain error, immediately after connection cleanup and before the redial-budget branch. Raise/throw `TunnelSupersededError` when the flag is set. Add regression tests where replacement `sendHello` sets the superseded flag and raises a plain reset/transient error, asserting exactly one attempt and no redial.

alex-w-99:
Confirmed and fixed in 2ba0c5d. You were right: `makeReplacementConnection` / `_make_replacement_connection` retried plain hello errors without consulting the takeover flag, so a superseded GOAWAY surfacing as a status-0 reset kept re-helloing for up to the 30s redial budget and repeatedly booted the winner. Both runtimes now check the terminal flag before retrying any plain error and stop immediately, matching the outer reconnect loop. Added Python + TS regressions (flag set + plain reset error -> exactly one attempt, no redial). Rust unaffected.

dimavrem22 (approved):
codex: Verified 2ba0c5da6012a8df6f9fe0b2ad81b3bc77c6039f addresses the handoff retry issue. The Python and TypeScript replacement loops now stop on the superseded flag before retrying plain hello errors, and the new regressions cover the flag-set/plain-error path. Local verification: Python superseded tests 20 passed, TypeScript superseded tests 9 passed, Rust tunnels-runtime tests 156 passed, focused repro now stops after one attempt. GitHub checks are green.

PR #96, opened 2026-07-06
Add date-range filtering to high-value comms list endpoints

dimavrem22:
Adds optional date-range filtering to the high-value comms list endpoints across all SDK surfaces: Python, TypeScript, Rust, and the CLI. Part of a coordinated change — companion PRs land in the servers and website repos (linked below).

## Contract

Three optional params are threaded onto each target list method (query params on the wire):

- `start_datetime` — string, default null
- `end_datetime` — string, default null
- `tz` — IANA zone (e.g. `America/New_York`), default null => UTC

All filtering is on the resource's `created_at` (stored UTC). The **server owns resolution** — the SDK/CLI only forward the raw strings, sending a param solely when it is non-null. Resolution semantics the server applies:

- `tz` governs interpretation of any zone-less value (null => UTC).
- Bare date `2026-07-01`: `start_datetime` => start of that day in `tz`; `end_datetime` => **whole-day inclusive** (all of the named day is returned, not truncated at 00:00).
- Datetime with explicit `Z`/offset => exact instant, `tz` ignored.
- Naive datetime (time, no zone) => interpreted in `tz`.
- Supplying only one side leaves the range open on the other.

## Non-breaking guarantee

- Omitting all three params is byte-for-byte identical to current behavior — no predicate is added.
- Ordering, pagination (cursor / offset / limit), and existing filters are untouched; the range only ANDs a `created_at` predicate server-side.
- New params are optional with null defaults, so existing SDK calls and tests keep working.

## Surfaces covered

Message list (+ unread), phone calls, texts, text conversations, iMessages, and iMessage conversations — at both the low-level resource layer and the `AgentIdentity` wrapper layer in each language, plus the corresponding CLI commands (`email list`, `email unread`, `phone calls`, `text list`, `text conversations`, `imessage list`, `imessage conversations`) via `--start-datetime` / `--end-datetime` / `--tz`.

## Notes

- Fully additive in all three languages. Python takes the params keyword-only, TypeScript as optional options fields, and Rust exposes them through a new `DateRangeFilter` struct passed to new `*_filtered` sibling methods (`list_filtered`, `list_conversations_filtered`, etc.) — the existing `list` / `iter_*` signatures are unchanged, so every existing caller compiles and behaves identically.
- All four packages bumped 0.4.19 -> 0.4.20 across every release surface: manifests, `CLI_VERSION` / TS `VERSION` constants, the CLI's `@inkbox/sdk` dependency pin, lockfiles, and the root + per-package changelogs.

---

## Companion PRs
Part of one cross-repo change (date-range filtering on comms list endpoints):
- Servers: inkbox-ai/servers#317
- Docs: inkbox-ai/website#130


dimavrem22:
Superseded by #101 (GROUPED FEATURES — merged to main). All commits from this PR are folded into that merge; closing to avoid a stale duplicate.

PR #98, opened 2026-07-08
`0.4.19`: Add email body fields to message.received webhooks

alex-w-99:
SDK + CLI support for the new email `body` fields on `message.received` webhooks.

- **Wire types** (Python / TypeScript / Rust): `data.message` gains `body`, `body_state` (`complete` / `truncated` / `unavailable`), `body_truncated`, `body_total_chars`, `body_included_chars`, and `email_address`. Mail context items gain `email_address`. All optional, so an older/replayed payload without them still parses.
- **Hydration:** when a body is truncated, fetch the full message with `messages.get(email_address, id)` (the row `id`, not `message_id`).
- Docs (READMEs + skills) and shared fixtures updated; tests cover the new fields and the old-payload case.
- Version bump to `0.4.19` across all four packages.


sdk/rust/src/webhooks/types.rs
```diff
@@ -390,6 +394,24 @@ pub struct MailWebhookMessage {
...
+    /// `messages.get(email_address, id)`. `default` tolerates pre-feature payloads.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub email_address: Option<String>,
+    /// Plain-text body on inbound `message.received` only (present-with-`null`
+    /// elsewhere). Whole under the size cap, else a prefix — see `body_state`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub body: Option<String>,
+    /// `complete` / `truncated` / `unavailable`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub body_state: Option<String>,
```
dimavrem22: codex: The Rust SDK loses the typed discriminator here: Python/TS expose `body_state` as `complete | truncated | unavailable`, and this file already models other webhook string discriminators as enums. Please add/export a `MailBodyState` enum with serde renames and type this as `Option<MailBodyState>` so Rust consumers can match exhaustively and invalid values do not deserialize silently.

alex-w-99: Good call, done in 54716b6. Added a `MailBodyState` enum (`#[serde(rename_all = "snake_case")]` -> `complete`/`truncated`/`unavailable`) and typed the field as `Option<MailBodyState>`, matching the other wire discriminators in this file and the Python/TS typings. It's re-exported via `pub use types::*`. Added a test asserting a valid variant round-trips and an unknown value fails to deserialize rather than parsing silently.

dimavrem22 (approved):
codex: Approved after re-reviewing the latest head. The Rust SDK now exposes MailBodyState as a serde snake_case enum, body_state is Option<MailBodyState>, unknown values fail to deserialize, and focused Rust/Python/TypeScript webhook type tests pass locally.

PR #100, opened 2026-07-09
SDKs: add blocked_spam_filter to SmsDeliveryStatus (0.4.20)

dimavrem22:
## What

The server persists `delivery_status=blocked_spam_filter` on outbound texts blocked pre-carrier by the Inkbox outbound spam filter, but no SDK enum included the value. Consequences today:

- **Python**: `texts.list(...)` / `texts.get(...)` over a conversation containing a blocked row raises `ValueError: 'blocked_spam_filter' is not a valid SmsDeliveryStatus` — the client crashes on data the server legitimately returns. (Found via a live CI failure in the Hermes plugin's delivery-retry suite, confirmed against production data.)
- **Rust**: same rows fail serde deserialization.
- **TypeScript**: no runtime failure, but the enum is incomplete.

## Change

- Add the variant to `SmsDeliveryStatus` in all three SDKs, documented as appearing on stored rows only — delivery webhooks never fire for blocked sends, so the webhook wire types (`SmsDeliveryStatusWire`) are intentionally unchanged.
- Python regression test: hydrating a spam-blocked row through `texts.list()`.
- Lockstep version bump 0.4.19 → 0.4.20 (`@inkbox/sdk`, `inkbox` PyPI, `inkbox` crates.io, `@inkbox/cli`) + CHANGELOG entry.

Once 0.4.20 is on PyPI, the hermes-agent-plugin live suite can raise its `inkbox>=` floor so its CI installs an enum-complete SDK (inkbox-ai/hermes-agent-plugin#45 currently works around this with a best-effort lookup).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
 Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
 (Python), `@inkbox/cli`, and `inkbox` (Rust, crates.io).
 
+## 0.4.20 — Recognize spam-filter-blocked texts
+
+### Fixed
+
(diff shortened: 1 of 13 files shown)


dimavrem22:
Superseded by #101 (GROUPED FEATURES — merged to main). All commits from this PR are folded into that merge; closing to avoid a stale duplicate.

PR #103, opened 2026-07-12
Rebrand prose: "hosted call agent" → "Inkbox Voice AI"

rayruizhiliao:
## What

Applies the Inkbox Voice AI rebrand (already live on the website docs) to every prose surface in this repo, and bumps all four packages to **0.4.23** so the rebrand can ship.

- **CLI**: `--help` strings in `phone.ts` (`--hosted`, `--reason`, `hosted-agent` subcommand descriptions) and the README command reference
- **Python SDK**: docstrings in `client.py`, `agent_identity.py`, `webhooks.py`, `phone/types.py`, `phone/resources/{calls,hosted_agent}.py`
- **TypeScript SDK**: JSDoc in `inkbox.ts`, `agent_identity.ts`, `phone/types.ts`, `phone/resources/{calls,hostedAgent}.ts`, `webhooks/types.ts`
- **Rust SDK**: doc comments in `client.rs`, `agent_identity.rs`, `phone/types.rs`, `phone/resources/{calls,hosted_agent}.rs`, `webhooks/types.rs`
- **Agent skills**: `inkbox-python`, `inkbox-ts`, `inkbox-cli` SKILL.md files, so agents learn the new vocabulary
- **Release prep**: lockstep 0.4.22 → 0.4.23 (pyproject, package.jsons + lockfile self-versions, Cargo.toml/Cargo.lock, CLI's `@inkbox/sdk` range) and 0.4.23 changelog sections in all four changelogs

Style matches the website docs: full "Inkbox Voice AI" on first/standalone mentions, "Voice AI" as the short form in follow-on references.

## Deliberately unchanged

- **API surface**: `mode="hosted_agent"`, `incoming_call_action="hosted_agent"`, the `inkbox phone hosted-agent` command, `HostedAgentConfig`-style type/method names, and file names — renaming any of these is a breaking change, and the docs pattern glosses them instead ("`hosted_agent` — Inkbox Voice AI answers")
- **Changelog history**: 0.4.22 and earlier entries keep the vocabulary they shipped with

## After merge

Publish per RELEASING.md (TS SDK before CLI): `publish.sh --prod` in `sdk/typescript`, `cli`, `sdk/python`, `sdk/rust`, then `git tag v0.4.23 && git push --tags`.

## Verification

- `grep -ri "hosted call agent\|hosted agent"` (excluding changelogs) returns zero prose hits
- Python: `py_compile` passes on all six edited files
- TypeScript SDK: `tsc --noEmit` clean
[...]

PR #104, opened 2026-07-12
Bump the two missed lockfile self-entries to 0.4.23

dimavrem22:
Stacked on #103. The 0.4.22 → 0.4.23 lockstep bump there missed two lockfile self-entries:

- `sdk/python/uv.lock` — the `inkbox` package's own entry (`source = { editable = "." }`)
- `cli/package-lock.json` — the linked `../sdk/typescript` (`@inkbox/sdk`) entry

Both now read 0.4.23; with this, no 0.4.22 remains anywhere outside changelog history.

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -26,7 +26,7 @@
     },
     "../sdk/typescript": {
       "name": "@inkbox/sdk",
-      "version": "0.4.22",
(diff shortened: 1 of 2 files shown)


dimavrem22:
Opened in error — passing these findings to the author instead.

PR #106, opened 2026-07-12
`0.4.23`: Mailbox storage, StorageLimitExceededError, mail clients

alex-w-99:
Adds mailbox storage reporting, a typed error for over-cap sends, and documentation for attaching a mail client over IMAP/SMTP.

## Mailbox storage

`Mailbox` gains `storage_used_bytes` and `storage_limit_bytes` (`storageUsedBytes` / `storageLimitBytes` in TypeScript) across the Python, TypeScript and Rust SDKs. Both are optional and defaulted, so a response without the fields parses to `0` / `null`.

Nothing is added to `Message` / `MessageDetail` — there is no per-message storage field on the wire.

## `StorageLimitExceededError`

Raised on a `402` whose `detail` carries `error = "storage_limit_exceeded"`, from `send`, `reply_all` and `forward` (and the `identity.*` delegators). Exposes `message`, `upgrade_url` and `limit_bytes`.

A `402` whose `detail` is a plain string stays a plain `InkboxAPIError`, so the SDKs behave sensibly against a server that doesn't send the structured body. There is no substring matching anywhere — the discriminator is the only thing consulted.

Deleting a message or thread frees its storage immediately, so `messages.delete` / `threads.delete` are the remedy the error points at.

## CLI

- `mailbox list` gains a `storage` column; `mailbox get` reports the raw byte counts. Humanized output uses **GiB/MiB** — the caps are binary, so a decimal label would name a different number of bytes. `--json` keeps the raw integers.
- The error handler prints how to free space or upgrade when a send is refused for storage.
- New `mailbox client-settings <email-address>`, which prints the IMAP/SMTP connection table for the configured API host. It never prints a password, and it errors rather than guessing if it can't map the host.

## Mail clients (IMAP/SMTP)

The READMEs and `skills/` now document attaching an inbox to a regular mail client: username is the inbox address, password is an identity-scoped API key (admin-scoped keys are rejected; one key maps to one mailbox). No new HTTP endpoint and no new SDK surface — the gateway speaks IMAP and SMTP directly.

The two constraints people actually hit are called out: `From` must be the authenticated inbox address, and the Free plan can't send signed or encrypted mail over SMTP. A client's own "save a copy of sent messages" is de-duplicated, so it can be left on.

## Testing

Python `pytest` 874 passed · TypeScript `vitest` 899 passed · CLI `node:test` 36 passed · Rust `cargo test` 138 passed.

New coverage includes the `402` discriminator and its plain-string fallback in all three SDKs, an omitted-fields (older response) parse case for the new mailbox fields, and the CLI's storage column and host-resolution behavior.

## Notes for review

- Version bumped to 0.4.23 in lockstep across all four manifests per `RELEASING.md`.
- **TypeScript is source-breaking**: `Mailbox` and `IdentityMailbox` gain required properties, so object literals in fixtures and mocks need updating. Parsing is unaffected.
- **Rust adds a `InkboxError` variant** and struct fields, which is source-breaking for exhaustive matches.


cli/src/commands/mailbox.ts
```diff
@@ -5,19 +5,115 @@ import {
...
+ */
+export function resolveMailDomain(baseUrl?: string): string | null {
+  if (!baseUrl) return INKBOX_MAIL_DOMAIN; // unset = the SDK's default API host
+  let host: string;
+  try {
+    host = new URL(baseUrl).hostname.toLowerCase();
+  } catch {
+    return null;
+  }
+  return INKBOX_API_HOSTS.has(host) ? INKBOX_MAIL_DOMAIN : null;
```
dimavrem22: codex: [high] This recognizes only production API hosts and always maps them to inkboxmail.com. With --base-url for development or beta, client-settings rejects a valid deployment even though servers #249 provisions the corresponding prefixed IMAP/SMTP hosts; map both CDN and direct API hostnames per environment and add tests for all three.

cli/src/commands/domain.ts
```diff
@@ -47,7 +47,7 @@ export function registerDomainCommands(program: Command): void {
     .command("set-default <domain-name>")
     .description(
       "Set the org default sending domain (admin-scoped API key only). " +
-        "Pass the platform sending domain (e.g. 'inkboxmail.com' in prod) to clear.",
+        "Pass the platform sending domain ('inkboxmail.com') to clear.",
```
dimavrem22: codex: [medium] This instruction is wrong outside production: clearing the default requires the configured environment platform domain, while inkboxmail.com is treated as a custom-domain lookup in beta/development. Restore environment-qualified wording or derive the platform domain from the configured base URL.

sdk/typescript/src/identities/types.ts
```diff
@@ -82,6 +82,13 @@ export interface IdentityMailbox {
   createdAt: Date;
   updatedAt: Date;
   filterModeChangeNotice: FilterModeChangeNotice | null;
+  /**
+   * Accepted for wire tolerance, but the identity embed does **not** report
+   * a real value today (reads `0`). Use `inkbox.mailboxes.get(...)`.
+   */
+  storageUsedBytes: number;
```
dimavrem22: codex: [data-model] The identity endpoint never serializes real storage values, so this public model knowingly exposes 0/null as if they were mailbox data and source-breaks TS/Rust struct literals for fields the endpoint cannot supply. Wire tolerance does not require public fields; keep storage on canonical Mailbox only, or first make the server identity serializer populate real values consistently.

sdk/rust/src/error.rs
```diff
@@ -68,6 +68,25 @@ pub enum InkboxError {
...
+    #[error("HTTP {status_code}: storage limit exceeded ({message})")]
+    StorageLimitExceeded {
+        status_code: u16,
+        /// Human-readable explanation from the server.
+        message: String,
+        /// Billing page where the plan can be upgraded.
+        upgrade_url: String,
+        /// The mailbox's storage cap. Binary bytes (Free is 2 GiB =
+        /// 2_147_483_648), so divide by 1024 and label GiB/MiB when displaying.
+        limit_bytes: u64,
```
dimavrem22: codex: [contract] Python/TypeScript preserve an absent limit_bytes as None/null, but Rust uses u64 and the parser silently converts absence/wrong type to 0. Zero cannot distinguish missing data and may be meaningful; use Option<u64> with missing-field coverage, or require/validate the field consistently across all SDKs.

cli/src/commands/mailbox.ts
alex-w-99: Fixed in [`8b3cbaa`](https://github.com/inkbox-ai/inkbox/commit/8b3cbaa2e043d76098c9ade5d762c8a5f818b9eb). `resolveMailDomain` now maps the supported production, beta, and development API hostnames to their corresponding mail domains. The CLI tests cover both recognized hostname forms for all three environments and retain the rejection test for unknown hosts.

cli/src/commands/domain.ts
alex-w-99: Fixed in [`8b3cbaa`](https://github.com/inkbox-ai/inkbox/commit/8b3cbaa2e043d76098c9ade5d762c8a5f818b9eb). Restored the environment-qualified wording so the help text presents `inkboxmail.com` specifically as the production example rather than as a universal value.

sdk/typescript/src/identities/types.ts
alex-w-99: Fixed in [`8b3cbaa`](https://github.com/inkbox-ai/inkbox/commit/8b3cbaa2e043d76098c9ade5d762c8a5f818b9eb). Removed the storage fields and parsing from `IdentityMailbox` in Python, TypeScript, and Rust. Storage remains on the canonical `Mailbox` model, and the changelogs and guides now describe that narrower surface.

sdk/rust/src/error.rs
alex-w-99: Fixed in [`8b3cbaa`](https://github.com/inkbox-ai/inkbox/commit/8b3cbaa2e043d76098c9ade5d762c8a5f818b9eb). Rust now exposes `limit_bytes` as `Option<u64>` and preserves missing or nonnumeric values as `None`. Regression coverage exercises both cases, while a valid numeric limit still parses as `Some(...)`.

dimavrem22 (approved):
Final Codex pass: all SDK/CLI contract findings are resolved on the latest head, cross-language models and environment mappings are consistent, and all checks pass.

PR #108, opened 2026-07-15
`0.5.0`: Identity tunnel summaries, tunnel ownership ids, inlined vault access

alex-w-99:
## Decisions

- **Identity list return types:** Extend existing summary types with optional hydrated channel and access fields. This preserves wire compatibility with older API responses, but adding public fields to Rust `AgentIdentitySummary` is source-breaking for struct literals and exhaustive patterns and is documented in the changelog.
- **Tunnel state:** Keep runtime state and certificate material on full tunnel reads.
- **Release scope:** Keep all SDKs and the CLI at version 0.5.0; publication remains a separate release step.

## Changes

- Python, TypeScript, and Rust identity list methods preserve mailbox, phone, iMessage, tunnel-summary, and access fields.
- Older summary-only list responses parse with empty channel and access defaults.
- Full tunnel responses expose `agent_identity_id`.
- Vault secret list/get responses expose inlined access rules.
- Public descriptions and fixtures use consumer-facing, provider-neutral wording.
- `inkbox tunnel get <handle>` resolves the summary and fetches the full tunnel.

## Verification

- Python focused suite: 166 passed; Ruff passed.
- TypeScript focused suite: 177 passed; build passed.
- Rust library suite: 164 passed; format and Clippy passed.
- Hosted Python, TypeScript, Rust, CLI, Development, and Beta checks pass on the current head.

## Release Gate

Version 0.5.0 is not published yet. Publish the TypeScript SDK before the CLI, publish all four packages in lockstep, then validate registry-backed installs before releasing dependent integrations.

PR #110, opened 2026-07-16
`0.4.25`: Proxy support in the CLI, clearer TypeScript connection errors

alex-w-99:
## Summary

Two independent fixes, released together as 0.4.25:

1. Node's `fetch` ignores `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` unless `NODE_USE_ENV_PROXY` is set, so in sandboxed or proxied environments — where many agents run — every CLI command and TS SDK request died with a bare `TypeError: fetch failed`. (Reported by a user running the CLI behind a sandbox proxy.)
2. Forward compatibility for the upcoming identity-tunnel contract change (#108): identity-embedded tunnel payloads will slim down to durable config, and today's 0.4.x parsers break on that shape.

## Changes

**CLI — proxies now just work.** When a proxy env var is present at startup, the CLI installs undici's `EnvHttpProxyAgent` as the global fetch dispatcher, so requests route through the configured proxy with `NO_PROXY` respected. If `NODE_USE_ENV_PROXY` is already set, the CLI defers to Node's own env-proxy handling. New runtime dependency: `undici` `^7` — v7 rather than v8 because v8 requires Node >= 22.19 while the CLI supports >= 22.

**TypeScript SDK — clearer connection errors.** New `InkboxConnectionError` (extends `InkboxError`) wraps any failure that occurs before an HTTP response exists (DNS, refused connection, TLS, unreachable proxy). The message names the request URL and the underlying cause, which Node buries in `error.cause`:

```
Error: Request to https://.../api/v1/whoami failed: getaddrinfo ENOTFOUND ...
```

When proxy env vars are set but Node isn't using them, the message appends a hint to run with `NODE_USE_ENV_PROXY=1` or configure a proxy-aware dispatcher. HTTP error responses (`InkboxAPIError` family) and timeouts (still an abort) are unchanged, and the original fetch error is preserved on `cause`.

**All three SDKs — tunnel fields tolerate omission.** The fields that #108's `TunnelSummary` drops from identity-embedded tunnels (`organization_id`, `cert_pem`, `cert_fingerprint_sha256`, `cert_expires_at`, `currently_connected`, `last_connected_at`, `last_connected_ip_addr`, `metadata`) are now parse-optional, so 0.4.25 clients survive the server-side change instead of breaking. Without this: TS coerces a missing `organization_id` to the string `"undefined"`, Python raises `KeyError`, and the Rust serde derive (the identity-embed path) rejects the payload outright.

- `organization_id` and `currently_connected` become nullable (`None`/`null`/`Option` = not reported — liveness is never fabricated as `false`).
- Cert and last-connected fields were already nullable; missing `metadata` still collapses to `{}`; unknown keys (e.g. the future `agent_identity_id`) are ignored.
- Source-breaking for Rust (`Option<String>` / `Option<bool>`) and strict-null TS consumers — called out in the changelogs, same convention as 0.4.24's `Mailbox` fields.

**Python and Rust SDKs — no proxy changes.** httpx and reqwest already honor proxy env vars by default.

**Release plumbing.** Lockstep bump to 0.4.25 (four manifests, four lockfiles, both User-Agent version constants, the CLI's `@inkbox/sdk` pin per RELEASING.md), changelog entries (root + per-package), and short proxy notes in the TS SDK and CLI READMEs.

## Decisions

- **Proxy fix in the SDK transport, not just the CLI**: direct SDK users hit the same bare `fetch failed`, and the CLI inherits the improved errors for free.
- **Hint-only in the SDK, full proxy support in the CLI**: the SDK stays zero-dependency (auto-proxying would require shipping undici to every consumer, including non-Node runtimes); the CLI owns its process and already has runtime deps, so it can fix the environment outright.
- After installing the dispatcher the CLI sets `NODE_USE_ENV_PROXY=1` so the SDK's hint can't misfire on CLI users whose proxy itself is down.
- **Tri-state over defaulting for `currently_connected`**: a fabricated `false` would misreport liveness; `null`/`None` says "not reported, fetch the tunnel by id for live state".

## Tests

- 4 new transport tests (error wrapping, cause fallback, hint on/off) in `sdk/typescript/tests/http.test.ts`.
- New `cli/tests/proxy.test.mjs`: a local TCP listener stands in for a proxy; the test runs the built CLI with `HTTPS_PROXY` pointing at it and asserts the connection arrives at the listener, proving requests are actually routed through the proxy.
- New durable-config-only tunnel parse tests in all three SDKs (TS `tunnels.test.ts`, Python `test_tunnels.py`, Rust inline in `tunnels/types.rs` covering both `from_value` and the serde-derive identity-embed path).

TypeScript 904 passed · Python 875 passed · CLI 37 passed · Rust 141 passed; clippy and rustfmt clean.


cli/src/index.ts
```diff
@@ -21,6 +22,15 @@ import { registerContactsCommands } from "./commands/contacts.js";
 import { registerNotesCommands } from "./commands/notes.js";
 import { registerDomainCommands } from "./commands/domain.js";
 
+// Node's fetch ignores HTTP(S)_PROXY/NO_PROXY unless NODE_USE_ENV_PROXY is
+// set, which strands the CLI in sandboxed/proxied environments. Honor them
+// ourselves, and set the flag so the SDK skips its proxy hint on errors.
+const proxyVars = ["HTTPS_PROXY", "https_proxy", "HTTP_PROXY", "http_proxy"];
+if (!process.env.NODE_USE_ENV_PROXY && proxyVars.some((name) => process.env[name])) {
```
dimavrem22: codex: This skips `EnvHttpProxyAgent` whenever `NODE_USE_ENV_PROXY` merely exists, but the CLI supports Node `>=22` while Node only added this setting in 22.21.0 ([docs](https://nodejs.org/download/release/latest-jod/docs/api/cli.html#node_use_env_proxy1)); on 22.20 and older, `NODE_USE_ENV_PROXY=1` leaves the default `Agent` installed and requests still bypass the proxy (and values such as `0` are also treated as enabled here). Please install the dispatcher on unsupported runtimes / disabled values, or always use it when proxy vars are present, and cover this supported-version case.

sdk/typescript/src/_http.ts
```diff
@@ -130,6 +143,21 @@ function raiseForErrorResponse(status: number, rawDetail: InkboxAPIErrorDetail):
...
 
+// Node's fetch ignores HTTP(S)_PROXY/NO_PROXY unless NODE_USE_ENV_PROXY is
+// set, so behind a mandatory proxy every request dies with a bare
+// "fetch failed". Point the user at the fix.
+function proxyHint(): string {
+  const env = typeof process === "undefined" ? undefined : process.env;
+  if (!env || env.NODE_USE_ENV_PROXY) return "";
+  const vars = ["HTTPS_PROXY", "https_proxy", "HTTP_PROXY", "http_proxy", "ALL_PROXY", "all_proxy"];
+  if (!vars.some((name) => env[name])) return "";
+  return (
```
dimavrem22, line 176: codex: This suggested fix is ineffective on Node 22.0–22.20, which this package supports: `NODE_USE_ENV_PROXY` was only added in 22.21.0. After following the hint, the next failure also hides it because line 149 treats any value as enabled; please tell these users to upgrade to 22.21+ or configure a dispatcher, and only suppress the hint when native env-proxy support can actually be active.

dimavrem22 (changes requested):
codex: Blocking on the supported-Node proxy compatibility issue in [this thread](https://github.com/inkbox-ai/inkbox/pull/110#discussion_r3598514869). The PR’s primary proxy fix still fails on supported Node 22.0–22.20 when `NODE_USE_ENV_PROXY` is present; please resolve that behavior and its SDK hint before merge.

cli/src/index.ts
alex-w-99: Fixed in 8dbc26f — the CLI now installs `EnvHttpProxyAgent` whenever proxy vars are present, regardless of `NODE_USE_ENV_PROXY`, so 22.0–22.20 (and 23.x) are covered. `NODE_USE_ENV_PROXY=0` opts out, matching Node's own semantics (verified empirically: Node dials direct under `=0`). Added tests for both: the flag-set case still routes through the fake proxy, and `=0` dials direct past it. Verified the version claim against the Node changelog — 22.21.0 (2025-10-20) is correct.

sdk/typescript/src/_http.ts
alex-w-99: Fixed in 8dbc26f — the hint now reads "run with NODE_USE_ENV_PROXY=1 (Node 22.21+ / 24+) or configure a proxy-aware fetch dispatcher (e.g. undici's EnvHttpProxyAgent) on older versions", so pre-22.21 users aren't sent down a dead end; test pins the caveat. I kept suppression env-based rather than version-sniffing: the CLI installs its own dispatcher on old Node and sets the flag as a marker, so "flag set but native support inactive" can mean proxying *is* active — version-gating the suppression would wrongly resurface the hint there. Anyone who set the flag themselves has now seen the version-caveated hint at least once.

dimavrem22: codex: Still open for standalone SDK users. Reproduced on Node 22.20 with `NODE_USE_ENV_PROXY=1` already present: a rejected fetch produces `Request to … failed: fetch failed.` with no version or dispatcher guidance, because line 149 returns before the new text; an environment may supply this flag before the SDK is ever run, so there is not necessarily an earlier caveated hint. Please distinguish native-supported/proxy-configured cases (for example via a dedicated CLI marker plus a Node-version check) before suppressing it.

alex-w-99: Fixed in 362cfd1, taking your suggested shape. The CLI now marks its installed dispatcher with a dedicated `INKBOX_ENV_PROXY_ACTIVE` instead of overloading `NODE_USE_ENV_PROXY`, and the SDK suppresses the hint only when env-proxying can plausibly be active: the marker is set, or the flag is set on a Node that honors it (`nodeSupportsEnvProxy`: 22.21+ within 22.x, or 24+), or the flag is an explicit `=0` opt-out. Your repro case — flag pre-baked on Node ≤22.20, standalone SDK — now gets a dedicated warning naming the running version ("NODE_USE_ENV_PROXY is set, but this Node version ignores it (… running 22.20.0)") instead of a bare `fetch failed`. Covered by unit tests on the version boundaries (22.20/22.21/23.x/24) plus branch tests for the marker, native-support, and flag-on-old-Node paths.

dimavrem22 (approved):
codex: Verified both blocking proxy-compatibility findings are resolved on this head. The CLI proxy path and standalone SDK hint now work on Node 22.20, version boundaries are covered, and all CI checks pass.

PR #118, opened 2026-07-19
CLI: support shared iMessage-line call origination

### LINKED ISSUE #91: CLI: support shared iMessage-line calls (--origination)
The SDKs can place shared iMessage-line calls (`origination=shared_imessage_number`, PR #90) but `inkbox phone call` has no `--origination` flag, so the CLI can't — the website docs dropped their CLI example because of this (inkbox-ai/website#124).

Done when: `inkbox phone call --to <e164> --origination shared_imessage_number` places a shared-line call via the identity's assignment.

CoderSufiyan:
## Problem

The SDKs can place calls through either an identity's dedicated number or its shared iMessage-line assignment, but `inkbox phone call` exposed only the default dedicated-number path. Identities without a dedicated number therefore could not place shared-line calls through the CLI even though the underlying SDK already supported them.

## Changes

- add `--origination <origin>` to `inkbox phone call`
- forward `dedicated_number` and `shared_imessage_number` to `AgentIdentity.placeCall()`
- preserve existing behavior when the option is omitted, allowing the SDK to default to `dedicated_number`
- include the server-returned `origin` in immediate call output; this makes shared calls understandable when `from` is null because the shared line is intentionally not exposed
- document the flag in the root quickstart, CLI README, and CLI agent skill
- document that shared origination needs an existing shared iMessage connection and otherwise returns `409 no_shared_connection`
- update the agent skill to confirm origination before placing a billable call

## Supported combinations

The option is forwarded independently of who drives the call, so it works with:

- the default client-driven call flow
- `--ws-url` audio bridging
- `--hosted --reason ...` Inkbox Voice AI calls

Example:

```bash
inkbox phone call -i my-agent --to +15551234567 \
  --origination shared_imessage_number
```

## Validation behavior

The CLI keeps its existing shape-only validation model. Hosted/reason/WebSocket conflicts still fail locally, while origination policy and unknown values remain server-validated and surface as normal API errors. This avoids duplicating server policy in the CLI.

## Tests

Added coverage for:

- omitted origination preserving the existing default path
- explicit `dedicated_number` forwarding
- `shared_imessage_number` forwarding
- shared origination with `--ws-url`
- shared origination with hosted Voice AI mode and reason
- existing hosted/reason/WebSocket conflict behavior

Verified locally with the repository's declared runtime:

- Node.js `v22.23.1`
- `npm test`: 42 passed, 0 failed
- TypeScript build: passed
- CLI `phone call --help`: new option rendered correctly
- `git diff --check`: passed

A live call was not placed because it requires a real shared assignment and creates a billable external side effect. Identity routing for shared origination is already covered by the SDK's `AgentIdentity.placeCall()` tests.

## Release scope

This PR does not bump package versions or changelogs. The repository releases Python, TypeScript, CLI, and Rust packages in lockstep, so release metadata is left to the coordinated maintainer release.

Closes #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,10 @@ inkbox email list -i my-agent --limit 10
 # Place a phone call
 inkbox phone call -i my-agent --to +15551234567
 
+# Or call over an existing shared iMessage connection (no dedicated number)
+inkbox phone call -i my-agent --to +15551234567 \
+  --origination shared_imessage_number
+
 # Send a text message (SMS/MMS; comma-separate --to for groups)
 inkbox text send -i my-agent --to +15551234567 --text "Hi from my agent!"
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -204,6 +204,8 @@ inkbox phone call -i <handle>                # Place an outbound call
   --hosted                                   #   Let Inkbox Voice AI drive the call
                                              #     (requires --reason; conflicts with --ws-url)
   --reason <text>                            #   Voice AI's task brief — what to accomplish
+  --origination <origin>                     #   dedicated_number (default) or
+                                             #     shared_imessage_number
 
 inkbox phone calls -i <handle>               # List calls
   --limit <n>                                #   Max results (default: 50)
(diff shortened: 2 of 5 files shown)


CoderSufiyan:
Following up on this PR. All CI checks are green, and the CLI coverage includes default, dedicated, shared, WebSocket, and hosted call paths. Happy to address any review feedback or adjust the scope if needed.

alex-w-99 (approved):

PR #122, opened 2026-07-22
`0.5.2`: Add contact memories to webhook payload types

alex-w-99:
## Decisions

- **Replay compatibility:** Keep `memories` optional in Python and TypeScript and default it to an empty vector in Rust so archived webhook replays remain consumable.
- **CLI scope:** Limit CLI behavior changes to the lockstep release bump because webhook bodies remain opaque to CLI verification and delivery commands.

## Changes

- Add newest-first contact `memories` arrays to mail, SMS, iMessage, incoming-call, and `call.ended` webhook types across all three SDKs.
- Update canonical fixtures, compatibility coverage, SDK documentation, and skills.
- Bump the Python, TypeScript, and Rust SDKs and CLI to 0.5.2; update the CLI dependency to `@inkbox/sdk` `^0.5.2`.

## Verification

- `uv run pytest` (915 passed, 3 skipped)
- `uv run ruff check .`
- `npm test` in the TypeScript SDK (938 passed, 3 skipped)
- `npm run build` in the TypeScript SDK
- `cargo test`
- `cargo test --features tunnels-runtime`
- `cargo clippy --all-targets --features tunnels-runtime -- -D warnings`
- `cargo fmt --check`
- `npm test` in the CLI (47 passed)


sdk/python/tests/test_webhook_types.py
```diff
@@ -211,12 +211,23 @@ def test_old_payload_without_body_fields_still_parses():
...
+
+
+def test_mail_contact_memories_and_old_replays():
+    payload = cast(MailWebhookPayload, _load("message_received.json"))
+    assert payload["data"]["contacts"][0]["memories"] == [
+        "Prefers email over phone calls."
+    ]
+    assert payload["data"]["contacts"][1]["memories"] == []
+    payload["data"]["contacts"][0].pop("memories")
+    assert "memories" not in payload["data"]["contacts"][0]
```
dimavrem22: nit: these two lines only exercise `dict.pop` — the assertion can't fail regardless of what the SDK declares. Since CI doesn't run a Python type-check step, there's no runtime way to verify `NotRequired` here; consider either dropping them or constructing a `WebhookMailContact` literal without `memories` (mirroring the TypeScript test) so a future type-check pass would cover it.

dimavrem22 (commented):
Reviewed against current server behavior: all five event families (mail, SMS, iMessage, incoming call, `call.ended`) emit `memories` on resolved contacts as active-only text, newest first, and `agent_identities` entries don't carry the field — the new types, fixtures, README/skill wording, and per-language replay handling all match. Lockstep bump, lockfiles, the `CLI_VERSION`/`VERSION` constants, the root changelog entry, and the Rust source-breaking note all follow the established conventions, and the untouched fixtures all have empty `contacts` arrays, so none of them went stale.

One follow-up outside this repo: the hosted webhook docs pages for mail and iMessage enumerate the contact entry fields and don't mention `memories` yet — worth a matching update once this ships.

sdk/python/tests/test_webhook_types.py
alex-w-99: Fixed in [`8f1278c`](https://github.com/inkbox-ai/inkbox/commit/8f1278c). The legacy shape is now a typed `WebhookMailContact` literal.

alex-w-99:
Hosted webhook docs follow-up is pushed across mail, phone/text/calls, and iMessage.

sdk/python/tests/test_webhook_types.py
dimavrem22: Confirmed — the typed literal covers the replay shape exactly as suggested. Thanks!

dimavrem22 (approved):
All feedback addressed and the full matrix is green — approving.

PR #125, opened 2026-07-24
`0.5.7`: Add mailbox import SDK and CLI support

alex-w-99:
## Decisions

- **Release version:** Ship the Python, TypeScript, and Rust SDKs plus the CLI together as `0.5.7` so every package exposes the same mailbox import contract.
- **Upload transport:** Upload archives directly to the returned target without API credentials. All SDKs expose bounded waits and upload timeouts; TypeScript additionally supports explicit upload cancellation.
- **Terminal states:** Return `completed`, `failed`, and `cancelled` jobs as typed results so callers can inspect counters and error details consistently.
- **API compatibility:** Mailbox import methods require an API deployment that exposes the matching import endpoints.

## Changes

- Add `mailboxes.imports` resources for create, upload, start, get, list, wait, refresh-upload-target, and cancel across all three SDKs.
- Add typed import jobs, formats, statuses, pagination, upload targets, and upload-specific errors.
- Add `inkbox mailbox imports run|get|list|wait|cancel`, upload progress, bounded command deadlines, and nonzero exits for failed or cancelled jobs.
- Expose `import_job_id` provenance as `importJobId` in TypeScript and update SDK guides, CLI docs, skills, integration coverage, and the changelog.
- Keep Python, TypeScript, Rust, and CLI package versions in lockstep at `0.5.7`; the CLI depends on `@inkbox/sdk` `^0.5.7`.

## Verification

- Python CI: `954 passed, 3 skipped`; lint passed.
- TypeScript CI: `979 passed, 3 skipped`; type checking passed.
- Rust CI: `200 passed` plus 2 doc tests with default features; `262 passed` plus 2 doc tests with tunnel features; builds, Clippy, and formatting passed.
- CLI local verification: build passed and `67 passed`.
- SDK integration checks are skipped while the PR is draft.
- Public-surface diff reviewed for internal references and credentials.


dimavrem22:
## Consistency review against servers #337

Reviewed this PR (`9c96f159`) against the servers implementation at `f46c8bcd`, checking every limit, enum, route, response field, error code and documented behavior. Timing matters here: this PR was last pushed **2026-07-26 19:57Z**, and servers has had three fix commits since (`eb54d27a`, `c4973400`, `f46c8bcd`) that changed the upload TTL, ZIP member handling, counter semantics and error shapes — so most of what follows is drift rather than anything wrong when it was written.

**The structural parity is genuinely good.** Routes, methods, status codes and every field of `MailImportJobResponse` / `MailImportUploadTarget` match across Python, TypeScript, Rust and the CLI — including optionality on `upload_size_bytes`, `error_detail` and `original_addresses`. All six statuses and four formats are present everywhere, no PUT-vs-POST breaks, and no SDK buffers a 1 GiB file in memory (Python streams a file handle, TS uses a file-backed `Blob`, Rust uses `Part::file`). All four manifests are consistent at `0.5.7` with lockfiles in step and a root `CHANGELOG.md` entry — convention satisfied.

---

### 1. The CLI never refreshes the 5-minute upload target — the flagship path breaks on large archives

`cli/src/commands/mailbox.ts:186`

Servers sets `MAIL_IMPORT_UPLOAD_URL_TTL_SECONDS = 300` while allowing `MAIL_IMPORT_MAX_UPLOAD_BYTES = 1 GiB`. `imports run` does create → upload → start with no re-issue and no retry:

```ts
const blob = await openAsBlob(file);
await inkbox.mailboxes.imports.upload(created.upload, blob, {...});
```

`grep -rn "refreshUploadTarget\|upload-url" cli/src/` returns **nothing** — the TS SDK exposes it, the CLI never calls it. Any transfer slower than ~28 Mbps sustained dies on an expired S3 policy, which is exactly the multi-hundred-MB Takeout archive this command exists for. Rust makes it worse: `sdk/rust/src/mail/resources/imports.rs:87` sets a `Duration::from_secs(3600)` client timeout, so it will sit for 55 minutes past policy expiry before giving up.

Worth catching the upload failure and re-issuing via `POST /{job_id}/upload-url` at least once.

### 2. A failed upload wedges the mailbox for 24 hours

`cli/src/commands/mailbox.ts:188`

`imports run` creates the job first, then uploads, with no `try`/`catch`. On an expired policy, a network drop or Ctrl-C the CLI exits leaving the row in `pending_upload` — which is in `ACTIVE_IMPORT_STATUSES`, so the one-active-import-per-mailbox unique index rejects the obvious user response (rerun the command) with `409 mail_import_already_in_flight` for the next **24 hours** (`MAIL_IMPORT_PENDING_UPLOAD_TTL_HOURS`). Each abandoned attempt also burns one of the 20 jobs/org/24h. `imports cancel` recovers it but is never suggested. Combined with finding 1, one flaky upload locks the mailbox out for a day.

### 3. "Counters may pause or reset during recovery" is now the opposite of the truth — 10 places

Servers changed this: counters are **cumulative across worker recovery** and never go backwards. This PR ships the old claim on every public surface:

```
README.md   cli/README.md   sdk/python/README.md   sdk/typescript/README.md   sdk/rust/README.md
skills/inkbox-all/SKILL.md   skills/inkbox-cli/SKILL.md   skills/inkbox-python/SKILL.md   skills/inkbox-ts/SKILL.md
```
plus `CHANGELOG.md:18`.

The servers wording is: *"Counters and checkpoints are cumulative across worker recovery and may remain unchanged while a slow message is processed."* This matters most in the **skills** — an agent told counters may reset will treat a stalled counter as normal churn rather than a problem, and won't trust `messages_imported` as a monotonic total.

### 4. The CLI prints raw JSON blobs for every import error

`cli/src/errors.ts:11`

```ts
return typeof detail === "string" ? detail : JSON.stringify(detail);
```

Servers now ships all five `mail_import_*` errors as `detail={"error": ..., "message": ...}`, and they all land in the generic `InkboxAPIError` branch. So the most common failure — retrying `imports run` after a failed upload — prints:

```
Error: HTTP 409: {"error":"mail_import_already_in_flight","message":"An import is already in flight for this mailbox."}
```

`StorageLimitExceededError` already shows the fix pattern at line 31 (`err.detailMessage ||`); the generic branch just needs the same treatment when `detail.message` is a string.

### 5. `Retry-After` on the 429 quota rejection is dropped in all three SDKs

Servers returns `429` with `detail={"error": "mail_import_quota_exceeded", ...}` **and** a `Retry-After` header. None of the three `_raise_for_status` implementations has a 429 branch (`sdk/python/inkbox/_http.py:241`, `sdk/typescript/src/_http.ts:203`, `sdk/rust/src/http.rs:357`), so callers get a bare error with no programmatic backoff hint. Rust even parses `retry_after_header` at `http.rs:339` but only consumes it in the 503 branch. The SDKs already have the pattern for this — `DedicatedIMessageNumberInventoryPendingError` captures `Retry-After` on 503.

### 6. "ZIP-of-EML" understates ZIP support — 11 files

Servers imports `.eml`, `.mbox`, `.mbx` **and** extensionless RFC 5322 members from a ZIP, specifically so a Google Takeout export (which contains a `.mbox`, not `.eml`) works unchanged:

```python
if lower_name.endswith((".mbox", ".mbx")):
    consumed, emitted = await self._consume_zip_mbox(archive, info)
```

Every surface here says "ZIP-of-EML", including the CLI `--help` text at `cli/src/commands/mailbox.ts:146`. A user holding the exact archive Takeout hands them is told to unzip and re-package first. Nothing states that non-mail and nested-archive members are **ignored** rather than failing the job, either.

### 7. Queueing semantics are undocumented, and it's the likeliest support question

Servers allows one RUNNING import **per organization** plus a 4-job cluster-wide ceiling, so a multi-mailbox org creates several jobs and watches them run strictly one at a time — and a job can sit `queued` behind a *different* organization entirely. None of the SDK/CLI/skill docs mention this. An agent following `skills/inkbox-all/SKILL.md:76` will read a long `queued` as a stall and cancel/recreate, burning quota and re-triggering the per-mailbox 409. Servers has the wording at `docs/mail-import.md:57-60`.

### 8. No limits documented on any public surface

Not one README, CLI doc or skill states any import limit. There are eight user-visible ones: 1 GiB upload, 50 MiB per message, 100,000 messages/job, 20 `original_addresses`, 20 jobs/org/24h, 24 h pending expiry, 300 s upload-URL TTL, 65,000 ZIP entries. Only the 1 GiB cap exists at all, as an un-documented CLI constant. `servers docs/mail-import.md:100-114` has a table this could mirror directly. Today a user discovers the 20/day quota or the 100k ceiling by getting a 429/422 in production.

---

### Smaller

- **`sdk/python/tests/test_mail_imports.py:45,114,132`** assert `expires_in_seconds=3600`, the retired value (servers returns 300). TS/Rust/CLI fixtures use 60, so no fixture in the repo reflects the real number — and these are exactly what gets copy-pasted into examples.
- **`sdk/rust/.../imports.rs:176`** — all three `wait()` timeout exits return `InkboxError::InvalidArgument`, the same variant used two lines above for `poll_interval must be greater than zero`. Python raises `TimeoutError` and TS throws a plain `Error`, so Rust is the only SDK where a wall-clock timeout can't be distinguished from caller misuse without string matching.
- **`sdk/rust/src/mail/types.rs:130`** — `MailImportJobStatus`/`MailImportFormat` have no `#[serde(other)]`, so a status the server adds later turns `get`/`list`/`wait` into a hard decode error rather than a degraded read. Matches how the rest of the file is written, so noting for awareness rather than as drift (Python shares the strictness; TS is tolerant via an unchecked cast).

### Verified clean

No vendor names (ClamAV/SES/Telnyx/Clerk), no internal repo names, no org IDs, no real addresses or phone numbers anywhere in the diff — all examples use `example.com`/`inkboxmail.com`. `timeout=3600` in the Python README and `Duration::from_secs(3600)` in the Rust README are `wait()` poll budgets, not the upload TTL, and are correct as written. Form fields correctly precede the `file` part in all three SDKs (Rust's `BTreeMap` reorders alphabetically, which is fine — S3 only requires `file` last). The integration tests exercise a real EML import end to end in all three lifecycle suites.

alex-w-99:
All eight findings and the three smaller items are addressed in fbe729655cc26fe357a28a23627852cb9f86723a. Verified each claim against current source before acting; the server-side revision you compared against has moved on since, which explains the one place your read and the code now differ.

Fixed:

1. `imports run` re-issues the upload target and retries the upload once, so a transfer that outlives the target no longer dies permanently.
2. On a still-failing upload, `run` cancels the job it created, and a 409 already-in-flight now prints an `imports list` / `imports cancel` recovery hint. An interrupted run no longer silently holds the mailbox.
3. Counter wording corrected everywhere: counters are cumulative and never go backwards, so a stalled counter is a signal rather than normal churn. Ten surfaces updated, skills included.
4. The CLI prints the server sentence for structured details instead of the raw JSON object.
5. `Retry-After` is captured on the 429 quota rejection in all three SDKs, and the CLI prints the retry hint.
6. ZIP wording now says a ZIP may hold `.eml` and/or `.mbox` members, that a Gmail Takeout ZIP imports as-is, and that non-mail and nested entries are ignored rather than failing the job. Eleven surfaces, CLI help included.
7. Queue behavior documented: jobs run one at a time per organization and share overall import capacity, so a long `queued` stretch is normal. The skills explicitly tell agents not to cancel and recreate.
8. Limits documented on every README, the CLI docs, and the skills: 1 GiB upload, 50 MiB per message, 100,000 messages and 20 original addresses per job, 65,000 ZIP entries, 20 jobs per organization per 24 hours, 5 minute upload target, 24 hour abandoned-upload expiry, one in-flight import per mailbox.

Smaller: the Python fixtures now use the real upload-target lifetime, and a Rust `wait()` timeout returns a dedicated variant instead of the caller-misuse one. Left the missing `#[serde(other)]` alone, since Python is equally strict and loosening both is a separate decision.

Source-breaking additions, all new API on a public SDK:

- Rust `InkboxError` gains `MailImportQuotaExceeded` and `Timeout` on top of the two upload variants already in this PR. Exhaustive matches must account for them, and import `wait()` now returns `Timeout` rather than `InvalidArgument` on a local timeout.
- Python gains `MailImportQuotaExceededError`, TypeScript gains `MailImportQuotaExceededError`. Both subclass the existing API error, so `except InkboxAPIError` / `instanceof InkboxAPIError` still catch them; code that asserted the base type exactly will see the new type.

The CHANGELOG entry for the unreleased 0.5.7 covers all of it, including the Rust note.

Parity gaps left deliberately:

- Upload auto-retry is CLI-only. The SDKs expose the refresh call and the docs tell callers to refresh-and-retry or cancel, but making retry the library default is a product call, not mine.
- The Rust 3600s upload timeout stays. It is a total-request budget for up to 1 GiB, not a hang: an expired policy returns an immediate rejection at request completion.
- Unknown-enum tolerance still diverges. TypeScript is tolerant via an unchecked cast, Python and Rust are strict.

Verification: Python 955 passed with ruff clean, TypeScript 980 passed with a clean build, Rust 202 passed with clippy and fmt clean, CLI 69 passed. New coverage: CLI retry-then-succeed and cancel-on-failure end to end, the 429 mapping in TypeScript, Rust, and Python, and Rust timeout versus misuse.

One thing to carry back to the backend: its import doc still describes ZIP as EML-only, which no longer matches the member handling the worker implements. The SDK docs here follow the code.

dimavrem22:
### Correction: my finding 1 rested on a premise I couldn't substantiate

Before you act further on the upload-window item, you should know I've withdrawn the equivalent finding on the console PR ([context](https://github.com/inkbox-ai/console/pull/213#issuecomment-5095815573)).

I claimed the 300 s presigned-POST TTL means a 1 GiB upload dies mid-transfer below ~28 Mbps. The counter-argument is that S3 evaluates a POST policy's `expiration` when the request *arrives*, not when the body finishes — the policy and signature are form fields parsed before `file`, so authentication happens up front and the 300 s bounds when the upload may *start*. That mechanism is plausible, and the only verbatim AWS wording I could retrieve points the same way:

> "Amazon S3 checks the expiration date and time of a signed URL at the time of the HTTP request. For example, if a client begins to download a large file immediately before the expiration time, the download continues even if the expiration time passes during the download."

That text is about presigned URLs rather than POST policies, so it isn't decisive — and I did find a secondary source claiming POST policies differ. But I couldn't verify that against AWS's own text, so I'm not going to keep asserting it.

**What this means for what you already shipped**, since you implemented against my premise:

- **Retry-once on a failed upload and cancel-the-job-on-failure are worth keeping regardless.** They don't depend on the TTL question at all. A transfer can fail for a dozen ordinary reasons, and before this a failed upload left the mailbox wedged in `pending_upload` for 24 h while burning a quota slot — that part of my finding 2 stands on its own and your fix is the right one.
- **The recovery hint on the 409** is likewise good independent of any of this.
- If the TTL genuinely gates arrival rather than completion, then the *re-issue* half of the retry is belt-and-braces rather than load-bearing. No harm in it, but don't let it drive further design.

Your note that the Rust 3600 s timeout is a total-request budget rather than a hang reads correctly to me on the same logic, so I withdraw the implied criticism there too.

It's cheaply settleable if you want certainty: presign with a 10-second TTL, push a file that takes 60 seconds through a throttled connection. I'd do it once, because the failure mode would be silent and would only appear on the large Takeout archives this feature exists for — but I'm not treating it as blocking.

The other seven findings and the three smaller items I'm verifying now and will come back on separately.

dimavrem22 (approved):
Approving. I re-derived every documented number against servers `src/constants.py` and `src/quotas/registry.py`, checked the prose against `docs/mail-import.md` at PR #337 head (including the changes made after my original review), and ran all four language surfaces.

**All eight findings and the three smaller items check out.**

- `imports run` genuinely re-issues and retries — `uploadWithRetry` calls the real `refreshUploadTarget` (POST `/{job_id}/upload-url`), with a test asserting `reissues=1, uploads=2, cancels=0`.
- Cancel-on-still-failing is legal from `pending_upload` (servers' `ACTIVE_IMPORT_STATUSES` includes it), and the 409 hint matches the exact code servers emits at `import_routes.py:212`.
- `"pause or reset"` and `"ZIP-of-EML"` are both **extinct repo-wide**.
- `Retry-After` is captured in all four surfaces — Python `_http.py:303`, TS `_http.ts:274`, Rust `http.rs:506` (reusing the already-parsed header), CLI prints the hint.
- Every limit re-derived and correct: 1 GiB / 50 MiB / 100,000 / 20 addresses / 65,000 ZIP entries / 20 jobs per org per 24 h / 300 s / 24 h.
- The four new typed errors subclass or extend `InkboxAPIError` correctly, are exported, and mirror `DedicatedIMessageNumberInventoryPendingError` field-for-field. `InkboxError` is genuinely not `#[non_exhaustive]`, so the source-breaking note is accurate rather than defensive. Versions 0.5.7 across all four manifests with lockfiles in step.

Tests: Python `955 passed` (88% coverage, ruff clean), TypeScript `980 passed` + `tsc --noEmit` clean, CLI `69 passed`, Rust `202 passed` with clippy and fmt clean.

---

**One thing I'd like fixed before merge**, because it quietly undoes item 7:

`skills/inkbox-python/SKILL.md:144`, `skills/inkbox-ts/SKILL.md:144` and `skills/inkbox-cli/SKILL.md:470` say counters are cumulative *"so a stalled counter is a signal, not normal churn"* — and stop there. `skills/inkbox-all/SKILL.md:85` keeps the caveat servers actually states:

> "Counters and checkpoints are cumulative across worker recovery and may remain unchanged while a slow message is processed."

Without that clause, "a stalled counter is a signal" is the exact reading that sends an agent to cancel and recreate a healthy job — the behaviour item 7 was added to prevent. All four READMEs got it right; only the three per-language skills lost it. One-clause edit to match `inkbox-all`.

**Follow-ups, none blocking:**

- `cli/src/commands/mailbox.ts:156` — `uploadWithRetry` uses a bare `catch {}`, so a deterministic rejection like the 413 `mail_import_upload_too_large` triggers a second full ~1 GiB transfer that cannot succeed. Narrowing to transport errors and expiry-shaped 403s keeps the fix and drops the wasted round trip. The first error is also swallowed, so only the second is ever shown.
- Auto-cancel frees the mailbox but the job still consumed one of the 20 daily slots, and there's no CLI way to re-issue against an existing `pending_upload` job (`refreshUploadTarget` is SDK-only) — so a transient blip costs a slot per attempt. A `--resume` in a follow-up, or a docs line saying a failed run consumes a slot.
- Items 2 and 4 (the 409 hint and the structured-detail rendering) are the two most likely to regress silently and neither is tested; both are cheap to cover with the existing `importRunServer` harness.
- TS and Rust fixtures still use `expires_in_seconds: 60`; only Python moved to the real 300.
- The root README and `cli/README.md` limits lists omit "one in-flight import per mailbox" — the limit users hit first, and the one the new 409 hint targets. The three SDK READMEs include it.
- `MailImportQuotaExceeded` exposes only `retry_after_header` in Rust while its sibling exposes a merged `retry_after_seconds` too. Equivalent today since servers sends it only as a header, but the accessor now differs by variant within one enum.
- Per-package CHANGELOGs still top out at 0.5.6. `RELEASING.md` §2 names a single root changelog so this is arguably in policy, but 0.5.4–0.5.6 all shipped per-package entries.

**On the declared gaps** — CLI-only auto-retry is the right call (implicit job cancellation would be surprising library behaviour). The Rust 3600 s upload timeout is fine as a total-request budget. The unknown-enum divergence is defensible for formats, but for `status` it isn't free: TS casts and `TERMINAL` is a `Set` of the six known values, so a future terminal status would fail `TERMINAL.has(...)` and `wait()` would poll to the caller's timeout instead of raising — Python and Rust both fail loudly there. Worth a note somewhere even if you keep the divergence.

Separately — your point about POST policy evaluation held up under a second look, which is the basis for the withdrawal I posted above.

PR #126, opened 2026-07-24
Prepare plugin manifests

alex-w-99:
## Decisions

- **Connection:** Bundle the remote MCP server with the existing skills.

## Changes

- Complete both plugin manifests.
- Add remote MCP configuration.
- Add strict manifest validation.
- Refresh installation docs.

## Verification

- Claude plugin validation
- Codex plugin validation

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -5,7 +5,7 @@
     "email": "hello@inkbox.ai"
   },
   "metadata": {
-    "description": "Inkbox email and phone SDK skills."
+    "description": "Inkbox communication tools and SDK skills."
   },
   "plugins": [
     {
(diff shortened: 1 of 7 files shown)


rayruizhiliao:
The catalog split is right, which is the thing that matters here: Claude plugin → `/mcp/anthropic`, Codex plugin → `/mcp`. The `version` fields and the `plugin validate --strict` CI job are good enforcement against manifest drift. One question + a nit:

- **`marketplace.json` dropped the `policy` block** (`installation: AVAILABLE`, `authentication: ON_INSTALL`). Confirm this was intentional — presumably either `--strict` rejects it or it's now superseded by MCP OAuth (the README's "Sign in when prompted"). Worth a one-line confirmation since it changes install-time auth behavior.
- **`skills/README.md` copy is internally inconsistent:** "Installs all six skills for automatic use, exposes self-signup as `/inkbox:inkbox-agent-self-signup`." The table has exactly 6 rows including `inkbox-agent-self-signup` — so self-signup can't be both one of the six auto-use skills and the separately-exposed slash command. Should read something like "six skills, five of which load automatically; self-signup is exposed as `/inkbox:...`".
- CI validates the `.claude-plugin` manifests but not `.codex-plugin` — expected (different schema, no Claude validator for it), just noting.

alex-w-99:
@rayruizhiliao Thanks. The `policy` removal is intentional: Claude strict validation rejects it, while Codex keeps the same defaults. Clarified the independent automatic/slash-command behavior in [05e11fe](https://github.com/inkbox-ai/inkbox/commit/05e11fe6d43027013dcee837d29bb3c4d574b4db). The Codex manifest also passes its official validator.

rayruizhiliao (approved):

PR #134, opened 2026-08-03
0.5.11: Make webhook contact names nullable

alex-w-99:
## Summary

`WebhookContact.name` and `WebhookMailContact.name` are now nullable in the Python, TypeScript, and Rust SDKs.

A contact created automatically from an inbound message carries an `id` and `memories` before anyone gives it a name. The API reports that name as `null` rather than repeating the phone number or email address the contact was matched on, so a receiver that reads `contact.name` gets an explicit "unknown" instead of an identifier that reads like a name.

All four packages move to **0.5.11** in lockstep.

## Why this matters most for Rust

Rust typed the field as `String`, so a payload with a null `name` failed to deserialize outright — the whole webhook, not just the field. Anyone handling inbound call, text, iMessage, or mail webhooks in Rust should upgrade.

TypeScript and Python typed it as non-nullable, which is a type-level inaccuracy rather than a runtime failure.

## Changes

| Package | Before | After |
|---|---|---|
| Rust | `pub name: String` | `pub name: Option<String>` (with `#[serde(default)]`) |
| TypeScript | `name: string` | `name: string \| null` |
| Python | `name: str` | `name: str \| None` |

Both READMEs document the nullability and the guidance not to fall back to the identifier. `CHANGELOG.md` carries the 0.5.11 entry.

## Compatibility

Source-breaking for Rust and TypeScript consumers that read `contact.name` as a non-optional value:

- Rust matches and struct literals must handle `None`.
- TypeScript needs a null check before use.
- Python's `TypedDict` widens, so type checkers flag the use site rather than the compiler.

Guard the field before addressing anyone by name. It deliberately does not fall back to the identifier: a missing name should read as unknown, not as a phone number.

## Verification

- Rust: `cargo fmt --check` clean, `cargo clippy --all-targets` clean, 214 tests + 2 doc-tests pass.
- Python: 975 passed, 3 skipped.
- TypeScript: `tsc --noEmit` clean, 996 passed, 3 skipped.
- CLI: 73 passed, 11 failing — the same 11 that fail on `main`, unrelated to this change.
- Lockstep version test passes across all four `package.json` / `pyproject.toml` / `Cargo.toml` files and their lock files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,33 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
 Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
 (Python), `@inkbox/cli`, and `inkbox` (Rust, crates.io).
 
+## 0.5.11 — Nullable webhook contact names
+
+### Changed
+
+- `WebhookContact.name` and `WebhookMailContact.name` are now nullable across
+  Python, TypeScript, and Rust. A contact created automatically from an inbound
+  message carries an `id` and `memories` before it has a name, and the field is
+  `None`/`null` in that case rather than repeating the phone number or email
+  address the contact was matched on.
+- Python, TypeScript, Rust, and CLI versions moved in lockstep to 0.5.11; the
+  CLI now depends on `@inkbox/sdk` `^0.5.11`.
+
+### Compatibility
+
+- **Source-breaking for Rust and TypeScript consumers that read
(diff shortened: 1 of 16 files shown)


dimavrem22 (approved):
Verified the latest head: the Python, TypeScript, and Rust webhook contact types match the explicit-null wire contract; Rust now accepts null names; package versions, lockfiles, and the CLI SDK dependency are consistent; and the package checks pass.

PR #137, opened 2026-08-04
0.5.13: Add A2A discovery and retire identity access

alex-w-99:
## Executive Summary

Ships A2A agent discovery and retires legacy identity visibility grants in release `0.5.13`.

- Adds typed public and organization directories with searchable cursor pagination.
- Adds public discoverability and public egress settings in Python, TypeScript, and the CLI.
- Removes identity-access types, response fields, methods, and CLI commands in favor of A2A directories and contact rules.

## Description

Python and TypeScript expose public and organization directory methods, iterators, typed directory entries, and the new A2A settings fields. Rust adds matching directory page types and resource methods. Platform Agent Card fetches authenticate only against the configured Inkbox origin and never send the Inkbox API key to external origins.

Release `0.5.13` removes `IdentityAccess`, identity visibility response fields, identity `list_access` / `grant_access` / `revoke_access` methods and their language-specific equivalents, plus the CLI `identity access` command group. Agent-scoped identity reads are self-only, and webhook `agent_identities` enrichment now matches active same-organization identities.

## Reason

Consumers need a supported way to discover eligible agents without exposing private identity channel configuration. Removing the overlapping visibility-grant contract gives discovery and communication policy one canonical A2A surface.

## Decisions

- **Credential scope:** Authenticate card reads only when the canonical origin matches the configured Inkbox origin; external card reads receive no Inkbox credential.
- **Rust scope:** Add directories to the existing Rust A2A resource without introducing a new protocol client or receiver-settings subsystem.
- **Compatibility:** Make the retirement source-breaking in `0.5.13` rather than retaining dead SDK wrappers around endpoints that now return `410 Gone`.
- **Versioning:** Move Python, TypeScript, Rust, and CLI packages in lockstep from `0.5.12` to `0.5.13`, with the CLI depending on `@inkbox/sdk` `^0.5.13`.

## Testing

- Python SDK: `978 passed`, `3 skipped`.
- TypeScript SDK: `1000 passed`, `3 skipped`.
- Rust SDK: `219 passed`; doc tests passed.
- CLI: `88 passed`.
- `cargo fmt --check` and Clippy with warnings denied passed.
- Version consistency checks for package manifests, lockfiles, SDK user-agent constants, and the CLI dependency passed.

## Backwards Compatibility

- [ ] Backwards compatible
- [x] NOT backwards compatible

`IdentityAccess`, identity visibility fields and methods, and the CLI `identity access` commands are removed. Consumers must migrate peer discovery to the A2A organization/public directories and communication controls to A2A contact rules before upgrading to `0.5.13`.


dimavrem22 (commented):
Preliminary review of the full diff at `d94ba29`, verified against the server-side contract. The engineering substance here is strong — the credential-scoping work on Agent Card fetches is correctly and defensively implemented in both implementing languages (exact canonical-origin equality, userinfo/HTTPS guards, redirect refusal, with negative tests), the wire contract matches the backend exactly on every param/key/default/termination detail checked, and the identity-access retirement leaves no stragglers across code, exports, docs, skills, or fixtures. The problems are concentrated in release mechanics:

### 1. [IMPORTANT] The PR double-bumps 0.5.12 → 0.5.14 and documents a phantom 0.5.13 that no commit can build

`CHANGELOG.md:7,28`. The root changelog adds both `## 0.5.14 — Retire identity visibility grants` and `## 0.5.13 — A2A agent discovery`, but main's last release is 0.5.12 and every manifest jumps straight to 0.5.14 in a single commit (at `2ca2bf6`, the "discovery" commit, manifests were still 0.5.12). The 0.5.13 entry's compatibility text — "Legacy identity-access methods remain available but are deprecated" — is not true of any buildable tree in this PR. A consumer reading the changelog could pin `inkbox==0.5.13` (never published; install fails) or expect deprecated-but-present access methods that exist in no shipped artifact. This also breaks the repo's one-patch-bump-per-release convention.

**Suggested fix:** collapse into a single `0.5.13` release entry (Added: directories + settings; Removed: identity access), or genuinely split into two releases where each version corresponds to a publishable commit.

### 2. [IMPORTANT] Per-package changelogs contradict the shipped manifests, and the source-breaking removals appear in no package changelog

`cli/CHANGELOG.md:12-13` says "CLI version moved in lockstep to 0.5.13 and now depends on `@inkbox/sdk` `^0.5.13`" while `cli/package.json` says `0.5.14` / `^0.5.14`. `sdk/python/CHANGELOG.md` and `sdk/typescript/CHANGELOG.md` likewise end at 0.5.13 with no entry for the removal of `IdentityAccess`/`listAccess`/the `identity access` command group — even though all three packages carry that breaking change. Bring these in line with whatever numbering #1 resolves to, and record the removals at the package level.

### 3. [IMPORTANT] The new pagination iterators have zero tests in either language

Python `iter_public_directory`/`iter_organization_directory` (`sdk/python/inkbox/a2a/resource.py:127-152`) and TS `iterPublicDirectory`/`iterOrganizationDirectory` (`sdk/typescript/src/a2a/resource.ts:137-160`) have no test references at all. The implementations read correct (cursor threaded, loop exits on falsy `next_cursor`), but a regression — e.g. dropping the `cursor = page.next_cursor` line — would ship as a silent infinite loop hammering the public endpoint. One multi-page mock test per language asserting (a) the page-2 request carries the cursor and (b) iteration stops on a null cursor.

### 4. [MINOR] Rust `A2ADirectoryVisibility` is a strict closed enum — a future visibility value fails the whole directory page only in Rust

`sdk/rust/src/a2a/types.rs:92-97` has no `#[serde(other)]`/`#[non_exhaustive]` fallback, while Python uses `ForwardCompatibleStrEnum` and TS uses an open string union. A new server-side visibility value would make `serde_json::from_value::<A2ADirectoryPage>` fail the entire listing call in Rust only. Add an `#[serde(other)] Unknown` variant or degrade to `String`.

### 5. [MINOR] Stale retired-wire-shape fixture

`cli/tests/a2a-command.test.mjs:199` — the identity mock still returns `"access": []`, a field no server version will return after this release. The test passes (nothing parses it), but it models a dead wire shape. Delete the line.

### 6. [MINOR] Rust rustdoc still promises "List all agent identities for your organisation" — now affirmatively wrong for agent-scoped keys

`sdk/rust/src/client.rs:526`. After the retirement, agent-scoped keys get only their own identity from list/get. An agent-scoped Rust consumer following the doc would iterate the list to find peers and find only itself, with no pointer to the organization directory. One line on the list/get docs in all three SDKs (and CLI `identity list` help) noting agent-scoped reads are self-only and pointing to `a2a` organization directory would close this.

### 7. [QUESTION] Are the tolerant defaults on the two new settings fields intended to be permanent?

`sdk/python/inkbox/a2a/resource.py:69-70,87-88` (`data.get("publicly_discoverable", False)`, `data.get("allow_public_egress", True)`) and the TS equivalents. They match server defaults and make sense for mixed rollout, but the server response makes both fields required — after rollout, tolerant parsing would silently mask a future server-side rename (a flag stuck reading `False` forever). Consider tightening to required parsing in a later release.

---

**Verified clean:** exact wire params `q`/`cursor`/`limit` and paths in all four surfaces; snake_case keys on the PUT body (test-asserted); limit default 50; iteration termination matches the server's null-cursor semantics; retirement grep-clean across sources, exports, READMEs, skills, and examples with removed-surface tests deleted (not skipped); 0.5.14 consistent across all manifests, lockfiles, `version.ts`, `CLI_VERSION`, and the CLI dep range, with the version-consistency test covering them; card-fetch credential scoping solid in Python and TS (canonical-origin `==`, `https` enforcement, userinfo rejection, `follow_redirects=False` + explicit 3xx refusal, external credential never carries the platform key) with negative tests in both; Rust's missing settings/protocol-client surface confirmed pre-existing scope, not new drift; no vendor names, real contact details, or internal references anywhere in the diff.

alex-w-99:
@dimavrem22 Thanks. Addressed the actionable findings in [`c923c6e`](https://github.com/inkbox-ai/inkbox/commit/c923c6e):

- Findings 1 and 2: collapsed discovery and retirement into one publishable `0.5.13` release across every manifest, lockfile, runtime constant, dependency range, and changelog.
- Finding 3: added two-page Python and TypeScript directory iterator tests that assert cursor progression, filter preservation, yielded order, and termination.
- Finding 4: Rust now tolerates unknown directory visibility values and has regression coverage.
- Finding 5: removed the retired `access` fixture field.
- Finding 6: corrected identity-list guidance across SDK READMEs, CLI help/docs, and Rust docs; agent-scoped callers are directed to the organization directory.
- Finding 7: retained the tolerant settings defaults intentionally for mixed-version compatibility; policy remains enforced by the server.

Verification: Python `978 passed, 3 skipped`; TypeScript `1000 passed, 3 skipped`; Rust `219 passed` plus doc tests; CLI `88 passed`; Rust fmt and Clippy passed.

dimavrem22:
Re-verified at `c923c6e` — all findings from the initial review are resolved:

1. ✅ Single `0.5.13` release: manifests, lockfiles, `VERSION`/`CLI_VERSION`, CLI dep `^0.5.13`, and README versions all agree; the phantom-release changelog split is gone.
2. ✅ Per-package changelogs now record the source-breaking removals and match the manifests.
3. ✅ Iterator tests added in both languages — `test_directory_iterators_follow_cursors_and_stop` and its TS counterpart assert the page-2 request carries the cursor and iteration stops on the null cursor, for both public and organization directories.
4. ✅ `A2ADirectoryVisibility` is now `#[non_exhaustive]` with `#[serde(other)] Unknown`, with a tolerance test.
5. ✅ Stale `"access": []` fixture removed.
6. ✅ Self-only list/get documented in the Rust rustdocs (both `client.rs` and the resource), CLI help, and changelogs, with pointers to the organization directory.

One remaining nit: the PR title and body still say `0.5.14` — worth updating both to `0.5.13` so the description matches the shipped release before this merges.

dimavrem22 (approved):
Approving — all review findings resolved as of `c923c6e`; wire contract, credential scoping, and release mechanics all verified. Update the PR title/body to 0.5.13 before merge.

PR #1, opened 2026-03-16
Add inkbox-kernel integration example

alex-w-99:
Adds the inkbox + Kernel (cloud browser) agent example under `examples/python/kernel/`. Gives an AI agent email, phone, and a browser via Inkbox and Kernel, with unified OpenAI/Anthropic support.

PR #143, opened 2026-08-08
Normalize SDK base_url (tolerate trailing /api or /api/v1)

rayruizhiliao:
## Problem

Both SDKs append `/api` and `/api/v1` to `base_url` themselves, so a caller who passes a **full** API base (e.g. `https://inkbox.ai/api`) gets doubled paths — `.../api/api/whoami` → **404**. This bit the agent onboarding flow end-to-end: the agent-signup response's own recommended `inkbox-claude bootstrap --base-url https://inkbox.ai/api` fails on its first call (`whoami`), and docs/humans routinely paste `/api` too.

`whoami` lives at the unversioned `/api/whoami` on purpose, so the SDK's `_api_base` is correct — the only bug is the un-normalized input.

## Change

- **Python** (`sdk/python/inkbox/client.py`): add `_normalize_base_url()` and apply it in the constructor. Strips a trailing `/api/v1` or `/api` (and trailing slashes) so callers may pass either the bare origin or a full API base.
- **TypeScript** (`sdk/typescript/src/inkbox.ts`): mirror with `normalizeBaseUrl()`.
- Tests added in both suites covering `/api`, `/api/v1`, and trailing-slash variants.

## Testing

- `uv run pytest tests/test_client.py` → 20 passed
- `vitest run tests/inkbox.test.ts` → 25 passed

## Related (separate PRs)

- servers: stop emitting `--base-url .../api` in the signup message (authoritative fix; this PR is the defensive half).
- claude-code-plugin: graceful voice skip, failing-step in bootstrap errors, real Claude-auth check in `doctor`.

Note: the in-repo `inkbox-agent-self-signup` skill already documents `invitation_token`; the stale coverage is only in the published v0.1.0 package (republish concern).

### DIFF
--- a/sdk/python/inkbox/client.py
+++ b/sdk/python/inkbox/client.py
@@ -69,6 +69,19 @@ from inkbox.whoami.types import WhoamiResponse, _parse_whoami
 
 _DEFAULT_BASE_URL = "https://inkbox.ai"
 
+
+def _normalize_base_url(base_url: str) -> str:
+    """Return the bare origin, tolerating a trailing ``/api`` or ``/api/v1``.
+
+    The SDK appends ``/api`` and ``/api/v1`` itself, so a caller who passes a
+    full API base (e.g. ``https://inkbox.ai/api``) would otherwise double it up.
+    """
+    trimmed = base_url.rstrip("/")
+    for suffix in ("/api/v1", "/api"):
+        if trimmed.endswith(suffix):
+            return trimmed[: -len(suffix)]
+    return trimmed
+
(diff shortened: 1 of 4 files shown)


rayruizhiliao:
Closing per decision to fix #1 server-side only (servers#480 now emits the bare origin `https://inkbox.ai`, which the current SDK resolves correctly). Keeping the SDK strict — base_url is expected to be the origin. Stale copies that still pass `/api` (published v0.1.0 CLI/skill, cached signup messages) will be corrected by the republish rather than absorbed defensively by the SDK.

PR #148, opened 2026-08-10
0.5.15: Harden tunnel runtime recovery

alex-w-99:
## Executive Summary

This PR makes long-running tunnel connections recover predictably across the Python, TypeScript, and Rust SDKs.

- Bounds connection establishment, HELLO, PING recovery, and task cleanup.
- Adds sampleable local liveness to Python and TypeScript listeners plus a Rust status handle.
- Ships deterministic recovery coverage, updated public guidance, and lockstep version 0.5.15.

## Description

Python and TypeScript retain make-before-break planned handoff while hardening cold recovery, waiter release, stale-response rejection, terminal error propagation, and lifecycle observability. Rust retains its blocking cold-reconnect model while adding interruptible establishment, explicit task ownership, bounded teardown, callback isolation, and a cloneable thread-safe status handle.

Status callbacks are edge-triggered across all three SDKs. Public READMEs, tunnel skills, package metadata, lockfiles, and the canonical changelog are updated; local `*PLAN.md` work artifacts are ignored.

## Reason

Transport loss or stalled establishment could strand tunnel work or delay recovery without a reliable local liveness signal. These changes make failure handling finite, observable, and consistently testable across SDKs.

## Decisions

- **Recovery ownership:** Keep one cold-retry owner per runtime so helpers fail and clean up without starting competing retry loops.
- **Planned drains:** Preserve Python and TypeScript make-before-break handoff and typed bridge drain semantics; keep Rust cold-reconnect only.
- **Liveness API:** Expose listener properties in Python and TypeScript, and use a cloneable status handle for Rust so its blocking API and caller-owned thread model remain unchanged.
- **Callback cadence:** Make lifecycle callbacks edge-triggered while retaining attempt-level detail in logs.
- **Compatibility:** TypeScript `serveForever()` now rejects terminal failures consistently with `wait()`; Python `close()` raises if its runtime thread exceeds the shutdown bound.

## Testing

- Python `TunnelListener` recovery and liveness: `uv run pytest` passed 1024 tests with 3 skipped; `uv run ruff check inkbox tests` passed.
- TypeScript `connect()`, `serveForever()`, and `wait()`: `npm test` passed 1031 tests with 3 skipped; build and bundle verification passed.
- Rust `TunnelRuntime` and `TunnelStatusHandle`: all-feature tests passed 309 tests; no-default-feature tests passed 231 tests; formatting and strict Clippy passed.
- CLI lockstep release metadata: `npm test` passed 106 tests.
- Complete public diff: whitespace, version consistency, and credential/internal-reference checks passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,34 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
 Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
 (Python), `@inkbox/cli`, `inkbox` (Rust, crates.io), and the bundled plugin.
 
+## 0.5.15 — Resilient tunnel runtimes
+
+### Added
+
+- Python and TypeScript tunnel listeners expose sampleable local `status`,
+  connection state, and latest successful connection time. Rust adds a
+  cloneable, thread-safe `TunnelStatusHandle` behind `tunnels-runtime`.
+
+### Changed
+
+- Python, TypeScript, and Rust bound tunnel connection establishment and HELLO,
+  clean partial connections and owned runtime tasks, and report cold reconnect
+  attempts and recovery. Missed PING acknowledgements force recovery at their
+  acknowledgement deadline.
+- Tunnel status callbacks are edge-triggered in all three SDKs: repeated failed
+  attempts within one outage retain `reconnecting` without emitting duplicate
+  callbacks. Attempt-level detail remains available in lifecycle logs.
+- Python and TypeScript preserve make-before-break planned handoff and typed
+  bridge drain while making cold connection loss release blocked work promptly.
+  Rust retains its blocking, caller-owned thread model and cold reconnect path.
+- TypeScript `serveForever()` now rejects on terminal runtime failures, matching
+  `wait()`. Fire-and-forget callers must attach a rejection handler.
+- Python `close()` raises `TimeoutError` rather than reporting success if its
+  runtime thread cannot stop within the shutdown bound. Signal and
+  `KeyboardInterrupt` cleanup remain non-raising.
+- Package and plugin versions moved in lockstep to 0.5.15; the CLI now depends
+  on `@inkbox/sdk` `^0.5.15`.
+
 ## 0.5.14 — A2A invitations
 
 - A2A invitation creates now preserve the server's optional share URL. Python
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
 - Python and TypeScript preserve make-before-break planned handoff and typed
   bridge drain while making cold connection loss release blocked work promptly.
   Rust retains its blocking, caller-owned thread model and cold reconnect path.
+- TypeScript in-process WebSocket handlers now receive typed
+  `WsConnectionLost` (`1011`) on cold tunnel loss; URL-forwarded WebSockets send
+  the same CLOSE code to their local upstream. Planned drain remains
+  `WsServerDraining` (`4500`).
 - TypeScript `serveForever()` now rejects on terminal runtime failures, matching
   `wait()`. Fire-and-forget callers must attach a rejection handler.
 - Python `close()` raises `TimeoutError` rather than reporting success if its
@@ -31,6 +35,15 @@ Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
   `KeyboardInterrupt` cleanup remain non-raising.
 - Package and plugin versions moved in lockstep to 0.5.15; the CLI now depends
   on `@inkbox/sdk` `^0.5.15`.
+
+### Compatibility
+
+- **Behavior change:** TypeScript callers that start `serveForever()` without
+  awaiting it must attach a rejection handler. Terminal failures no longer wait
+  for a later `wait()` call to surface.
+- **Behavior change:** Python `close()` can raise `TimeoutError` after 30 seconds.
+  A `finally` block should preserve any earlier exception if shutdown also times
+  out. Signal and `KeyboardInterrupt` cleanup remain non-raising.
 
 ## 0.5.14 — A2A invitations
 
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ use inkbox::tunnels::client::TunnelStatusHandle;
 let status = TunnelStatusHandle::new();
 let runtime_status = status.clone();
 let client = inkbox.clone();
-std::thread::spawn(move || {
+let tunnel_thread = std::thread::spawn(move || {
     client.tunnels().connect_with_status(
         "my-app",
         "http://127.0.0.1:8080",
@@ -280,6 +280,11 @@ std::thread::spawn(move || {
     )
 });
 println!("{:?} {}", status.status(), status.is_connected());
+
+// Join when shutdown is expected so startup/runtime errors are surfaced.
+if let Err(error) = tunnel_thread.join().expect("tunnel thread panicked") {
+    eprintln!("tunnel stopped: {error}");
+}
 ```
 
 Python and TypeScript also accept an in-process callable (Fetch handler in TS,
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ Task detail includes current state and message history.
 # Outbound HTTP/2 only — no inbound port to open. POSIX only.
 listener = inkbox.tunnels.connect(name="my-app", forward_to="http://127.0.0.1:8080")
 print(listener.public_url)
+print(listener.status, listener.is_connected, listener.last_connected_at)
 listener.wait()
 ```
 
@@ -256,10 +257,35 @@ const listener = await connect(inkbox, {
   forwardTo: "http://127.0.0.1:8080",
 });
 console.log(listener.publicUrl);
+console.log(listener.status, listener.isConnected, listener.lastConnectedAt);
 await listener.wait();
 ```
 
-Both SDKs also accept an in-process callable (Fetch handler in TS, ASGI app in Python) instead of a `forward_to` URL, and a `tls_mode: "passthrough"` option for end-to-end TLS termination in your process. See [`skills/inkbox-tunnels/`](./skills/inkbox-tunnels/) for the full reference.
+### Tunnels (Rust)
+
+The blocking Rust runtime is behind the `tunnels-runtime` feature. Run it on a
+caller-owned thread and retain a cloneable local status handle:
+
+```rust
+use inkbox::tunnels::client::TunnelStatusHandle;
+
+let status = TunnelStatusHandle::new();
+let runtime_status = status.clone();
+let client = inkbox.clone();
+std::thread::spawn(move || {
+    client.tunnels().connect_with_status(
+        "my-app",
+        "http://127.0.0.1:8080",
+        runtime_status.callback(),
+    )
+});
+println!("{:?} {}", status.status(), status.is_connected());
+```
+
+Python and TypeScript also accept an in-process callable (Fetch handler in TS,
+ASGI app in Python) instead of a `forward_to` URL. All three runtimes reconnect
+after transient connection failures. See [`skills/inkbox-tunnels/`](./skills/inkbox-tunnels/)
+for the full reference.
 
 **Redeploys are graceful.** When the tunnel service redeploys, a long-running listener reconnects make-before-break: it stands up a fresh connection before closing the draining one, so short HTTP requests see no gap. In-progress WebSocket and passthrough-TCP sessions cannot migrate across a redeploy — they end with a typed `server_draining` close and the third-party peer reconnects onto the new task. Write handlers to reconnect idempotently.
 
--- a/sdk/python/inkbox/tunnels/client/_listener.py
+++ b/sdk/python/inkbox/tunnels/client/_listener.py
@@ -150,7 +150,11 @@ class TunnelListener:
             raise err
 
     def close(self) -> None:
-        """Sync graceful shutdown."""
+        """Sync graceful shutdown.
+
+        Raises:
+            TimeoutError: If the runtime thread does not stop within 30 seconds.
+        """
         self._close(raise_on_timeout=True)
 
     def _close(self, *, raise_on_timeout: bool) -> None:
--- a/sdk/python/inkbox/tunnels/client/_listener.py
+++ b/sdk/python/inkbox/tunnels/client/_listener.py
@@ -17,6 +17,7 @@ import logging
 import signal
 import sys
 import threading
+from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
@@ -30,6 +31,7 @@ from inkbox.tunnels.client._runtime import (
     DEFAULT_OUTBOUND_BODY_BYTES,
     StatusCallback,
     TunnelRuntime,
+    TunnelRuntimeStatus,
 )
 from inkbox.tunnels.client._url_forward import validate_forward_target
 from inkbox.tunnels.types import Tunnel
@@ -62,6 +64,9 @@ class TunnelListener:
         tunnel: A snapshot of the :class:`Tunnel` resource record taken
             at bootstrap. Not refreshed; call ``inkbox.tunnels.get(id)``
             for live state.
+        status: Current local runtime state.
+        is_connected: Whether the local runtime is currently connected.
+        last_connected_at: UTC time of the latest successful connection.
     """
 
     def __init__(
@@ -91,6 +96,18 @@ class TunnelListener:
     def tunnel(self) -> Tunnel:
         return self._bundle.tunnel
 
+    @property
+    def status(self) -> TunnelRuntimeStatus:
+        return self._runtime.status
+
+    @property
+    def is_connected(self) -> bool:
+        return self._runtime.is_connected
+
+    @property
+    def last_connected_at(self) -> datetime | None:
+        return self._runtime.last_connected_at
+
     # --- sync API -----------------------------------------------------------
 
     def wait(self) -> None:
@@ -115,7 +132,7 @@ class TunnelListener:
                 while not self._stopped.wait(timeout=1.0):
                     pass
             except KeyboardInterrupt:
-                self.close()
+                self._close(raise_on_timeout=False)
                 raise
         finally:
             if sigterm_handler_installed:
@@ -134,11 +151,22 @@ class TunnelListener:
 
     def close(self) -> None:
         """Sync graceful shutdown."""
+        self._close(raise_on_timeout=True)
+
+    def _close(self, *, raise_on_timeout: bool) -> None:
         loop = self._loop
         if loop is not None and not loop.is_closed():
             loop.call_soon_threadsafe(self._schedule_async_close)
         if self._thread is not None:
             self._thread.join(timeout=30.0)
+            if self._thread.is_alive():
+                message = "tunnel runtime thread did not stop within 30 seconds"
+                if raise_on_timeout:
+                    raise TimeoutError(message)
+                logger.error(message)
+        elif self._runtime.status != "closed":
+            self._runtime._stop.set()
+            self._runtime._notify_status("closed")
 
     # --- async API ----------------------------------------------------------
 
@@ -173,7 +201,7 @@ class TunnelListener:
 
     def _signal_handler(self, signum: int, frame: object) -> None:
         logger.info("received signal %s; shutting down listener", signum)
-        self.close()
+        self._close(raise_on_timeout=False)
 
     def _start_thread_if_needed(self) -> None:
         if self._thread is not None:
@@ -256,7 +284,7 @@ def connect(
             (1-32). Omit to let the server decide.
         on_status: Callback invoked with status strings
             (``"connecting"``, ``"connected"``, ``"reconnecting"``,
-            ``"closed"``).
+            ``"closed"``, ``"superseded"``).
         max_inbound_body_bytes: Cap on materialized inbound bodies;
             oversize requests get a 413 to the third party.
         max_outbound_body_bytes: Cap on materialized outbound bodies;
--- a/sdk/python/inkbox/tunnels/client/_runtime.py
+++ b/sdk/python/inkbox/tunnels/client/_runtime.py
@@ -28,9 +28,12 @@ import random
 import socket
 import ssl
 import struct
+import threading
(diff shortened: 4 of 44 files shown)


sdk/python/inkbox/tunnels/client/_runtime.py
```diff
@@ -549,12 +640,16 @@ async def _run_once(self) -> None:
...
+                if self._handoff_terminal_error is not None:
+                    raise self._handoff_terminal_error
                 nxt = self._active
                 if nxt is not None and nxt is not conn and not nxt.draining:
                     conn = nxt
                     continue
                 if conn.read_task is None or conn.read_task.done():
                     break
+            if conn.read_task is not None and conn.read_task.done():
+                conn.read_task.result()
```
dimavrem22: **Bug (blocking): a failed handoff terminates the runtime instead of falling back to cold reconnect.**

When a handoff fails, `_run_handoff` forces the old conn down and `_drain_old_connection` cancels its read task; the cancellation reliably wins over the transport-abort wakeup, so the supervised read task finishes **cancelled** and this unconditional `result()` raises `CancelledError`. `serve_forever`'s `except asyncio.CancelledError: raise` then propagates it — no cold redial, contradicting the comment in `_run_handoff` that says the supervisor will reconnect cold. Worse, the sync listener's `_runner` swallows `CancelledError`, so `wait()` returns cleanly: the tunnel dies silently, with no error and no reconnect, in exactly the scenario this PR hardens (e.g. server drain plus a bad 30s redial window).

Reproduced with a focused test (`serve_forever` ended in `CancelledError`, zero cold reattempts). Guarding with `and not conn.read_task.cancelled()` fixes the repro with the full tunnel suite still green.

Note the same line also means a normal `aclose()` during a live connection now ends `serve_forever` via `CancelledError` instead of a clean return — a behavior change for anyone awaiting `serve_forever()` directly.

There's no test for the plain handoff-failure → cold-reconnect path (the handoff-auth test replaces `_wait_close_or_handoff` with a custom driver, so real supervisor sequencing is never exercised) — please add one alongside the fix.

sdk/python/inkbox/tunnels/client/_runtime.py
```diff
@@ -674,34 +837,61 @@ async def _run_handoff(self, old_conn: _Connection) -> None:
             self._supervisor_wake.set()
         finally:
             self._handoff_in_flight = False
-            self._handoff_task = None
-            await self._drain_old_connection(old_conn)
+            try:
+                await self._drain_old_connection(old_conn)
+            finally:
+                self._handoff_task = None
```
dimavrem22: Race introduced by the ordering here: `_handoff_in_flight` is cleared *before* the drain, but `self._handoff_task = None` runs unconditionally *after* it. If a second handoff starts while handoff A is still draining (a second NO_ERROR GOAWAY on the new active conn), `_begin_handoff` sets `_handoff_task` to B and A's `finally` then nulls it. `aclose()` snapshots `_handoff_task` and so misses B — B can complete a replacement dial concurrently with shutdown and install a fresh active conn whose writer/read/intake tasks are never torn down until B's loop happens to notice `_stop`. The pre-change code nulled the reference before the drain, so there was no window.

Suggest compare-and-clear (`if self._handoff_task is asyncio.current_task(): self._handoff_task = None`), and/or having `_begin_handoff` refuse to start while `_handoff_task` is still set.

sdk/python/inkbox/tunnels/client/_listener.py
```diff
@@ -134,11 +151,22 @@ def wait(self) -> None:
 
     def close(self) -> None:
         """Sync graceful shutdown."""
```
dimavrem22: `close()` raising `TimeoutError` is a real public behavior change — a `finally: listener.close()` can now mask the original exception — but the docstring still just says "Sync graceful shutdown." Please document the raise here (the changelog mentions it; the Python README's new tunnels section doesn't). Also, the 30s join bound below is a hardcoded magic number while every other bound in this PR is a named constant — worth hoisting.

sdk/typescript/src/tunnels/client/_runtime.ts
```diff
@@ -372,11 +444,19 @@ export class TunnelRuntime {
...
+          );
+        } else {
+          // eslint-disable-next-line no-console
+          console.warn(
+            `tunnel runtime: connection error on attempt #${attempt} ` +
+              `(consecutive failure #${consecutiveFailures}); reconnecting`,
+            err,
+          );
+        }
         this.notifyStatus("reconnecting");
```
dimavrem22, line 468: This emission has no `!this.stop` guard (the equivalent one in `runOnce` does). `aclose()` during an in-flight dial/HELLO emits `closed` first, then the teardown-induced dial failure lands here — observers see `closed → reconnecting → closed`, and `listener.status` can transiently read `reconnecting` after close began. One-line fix: guard with `!this.stop`.

sdk/typescript/src/tunnels/client/_runtime.ts
```diff
@@ -2221,7 +2438,11 @@ export class TunnelRuntime {
   /** The active conn if it can take new streams, else the origin if it can. */
   private pickReplyConnection(origin: Connection): Connection | null {
     const usable = (c: Connection | null): c is Connection =>
-      c !== null && !c.draining && c.session !== null && !c.session.closed;
+      c !== null &&
+      !c.draining &&
+      c.session !== null &&
+      !c.session.closed &&
+      !c.session.destroyed;
```
dimavrem22, line 2459: Cross-SDK gap: Python and Rust both gained a stale-reply generation fence in this PR (Python stamps dispatch tasks with a generation ContextVar and `_post_response` drops mismatches; Rust rejects in `post_response` when the connection generation moved). TypeScript only checks closed/destroyed/draining here — a dispatch continuation that survives a cold reconnect can still migrate its reply onto the **new** session, which the other two SDKs now explicitly drop. Python has four tests for this; TS has neither the mechanism nor tests. If the omission is intentional it deserves a note; otherwise TS should get the same generation fence.

sdk/typescript/src/tunnels/client/_runtime.ts
```diff
@@ -436,52 +522,132 @@ export class TunnelRuntime {
...
       }
     });
   }
 
+  private publishConnectionClosed(conn: Connection): void {
+    if (!conn.closePublished) {
+      conn.closePublished = true;
+      for (const [, bus] of conn.streams) {
+        if (!bus.ended) {
+          bus.events.push({ kind: "reset", code: 0 });
```
dimavrem22, line 574: Cross-SDK divergence, customer-visible: on cold connection loss, Python bridges now close customer WebSocket streams with a real close code (1011, or the planned 4500/1000 drain codes via `bridge_close_code`), but TS still publishes `{ kind: "reset", code: 0 }` for every stream. A customer's WS handler sees different close semantics depending on SDK language. Worth aligning with the Python behavior (or documenting the divergence in the changelog).

sdk/typescript/src/tunnels/client/_listener.ts
```diff
@@ -154,7 +162,7 @@ export class TunnelListenerImpl implements TunnelListener {
       try {
         await this.servePromise;
```
dimavrem22: Narrow crash window in the new rejection semantics: this swallowing `catch` only attaches after `await this.runtime.aclose()`, which crosses real timers (session-close grace ~250ms). A terminal rejection (takeover/auth) that settles in that window still fires `unhandledRejection` — even for a caller doing the documented close dance. Attaching a passive handler on a branch at creation time in `serveForever()` (e.g. `void promise.catch(() => {})` while returning the unbranched promise) would remove the crash window; note it also means fire-and-forget callers never crash, which softens the documented "must attach a rejection handler" contract — either way it should be a deliberate choice rather than an accident of ordering.

sdk/rust/src/tunnels/client/runtime.rs
```diff
@@ -494,15 +702,22 @@ impl TunnelRuntime {
...
                     let me = self.clone();
                     let c = conn.clone();
-                    tokio::spawn(async move {
+                    let handle = tokio::spawn(async move {
                         let _ = me.dispatch(env, c).await;
                     });
+                    dispatch_handles
+                        .lock()
+                        .unwrap_or_else(|err| err.into_inner())
+                        .push(OwnedTask::new("request dispatch", handle));
```
dimavrem22: **Bug (blocking): unbounded accumulation on a healthy connection.** Every dispatched request pushes an `OwnedTask` here, and nothing removes entries until connection teardown drains the vec. A held `JoinHandle` keeps the finished task's allocation alive, so on a stable long-lived tunnel under steady traffic, memory grows linearly with request count — the healthy steady state this PR is meant to protect. Prune finished handles on push (`retain(|t| !t.handle.is_finished())`) or use a `JoinSet` and reap completed tasks.

sdk/rust/src/tunnels/client/runtime.rs
```diff
@@ -768,13 +986,217 @@ impl TunnelRuntime {
...
+                .lock()
+                .unwrap_or_else(|err| err.into_inner());
+            if last.as_deref() == Some(status) {
+                return;
+            }
+            *last = Some(status.to_string());
+        }
         if let Some(cb) = &self.cfg.on_status {
-            cb(status);
+            if catch_unwind(AssertUnwindSafe(|| cb(status))).is_err() {
```
dimavrem22, line 1101: Two issues with callback delivery here:

1. **Panic loses the edge permanently.** `last_notified_status` is recorded *before* the callback runs, so a panicking callback leaves the transition marked delivered — edge-triggering suppresses re-emission and a composed `TunnelStatusHandle` stays stale on the previous state until a *different* status arrives. Consider recording the status only after the callback returns without panicking (the dedupe check can stay first).
2. **Blocking callbacks can stall the data plane.** The callback runs inline on a runtime worker, and the design explicitly tolerates blocking callbacks — but `connect()` builds the runtime with default worker count = num CPUs, so on a 1-vCPU host a blocking `connected` callback freezes intake, the PING loop, and the supervisor for its duration (long enough → missed pings → the server drops the connection). The test that proves "blocking is safe" pins `worker_threads = 4`. Either run callbacks via `spawn_blocking` or document that callbacks must not block.

sdk/rust/src/tunnels/client/runtime_tests.rs
```diff
@@ -0,0 +1,864 @@
...
+    tokio::time::timeout(Duration::from_millis(500), serving)
+        .await
+        .unwrap()
+        .unwrap()
+        .unwrap();
+    signals.abort_tasks();
+}
+
+#[tokio::test]
+async fn ping_timeout_forces_reconnect_before_another_interval() {
```
dimavrem22, line 641: Flake risk: this test asserts an exact 5-status sequence plus a <320ms wall-clock bound against a ~230ms nominal path. If `aclose` is delayed more than ~230ms past the second `connected` on a loaded CI runner, the second connection's ping cycle (200ms interval + 30ms ack) can inject an extra `reconnecting` and break the exact-sequence assert. Several sibling tests assert <100–150ms post-`aclose` margins too. `start_paused` virtual time is already used for `phase_deadline` in this file — consider it here, or loosen the exact-sequence assert to a subsequence check.

README.md
```diff
@@ -256,10 +257,35 @@ const listener = await connect(inkbox, {
...
+
+The blocking Rust runtime is behind the `tunnels-runtime` feature. Run it on a
+caller-owned thread and retain a cloneable local status handle:
+
+```rust
+use inkbox::tunnels::client::TunnelStatusHandle;
+
+let status = TunnelStatusHandle::new();
+let runtime_status = status.clone();
+let client = inkbox.clone();
```
dimavrem22, line 274: This example doesn't compile: `Inkbox` has no `Clone` impl anywhere in the crate, so `inkbox.clone()` is an error. The same pattern appears in `sdk/rust/README.md` and `skills/inkbox-tunnels/SKILL.md`. Either `Inkbox` was meant to gain `Clone` in this PR, or the examples should move `inkbox` into the thread (or wrap it in `Arc`).

sdk/rust/README.md
```diff
@@ -388,10 +388,38 @@ pure Rust.
...
+The connect methods remain blocking. Run them on a caller-owned thread and use
+`TunnelStatusHandle` to sample local liveness elsewhere:
+
+```rust
+use inkbox::tunnels::client::TunnelStatusHandle;
+
+let status = TunnelStatusHandle::new();
+let runtime_status = status.clone();
+let client = inkbox.clone();
+std::thread::spawn(move || {
```
dimavrem22: Beyond the `inkbox.clone()` compile error (flagged on the root README): this example discards the spawned thread's `Result`, and `connect_with_status` does all of its fallible bootstrap (name validation, REST calls, CSR signing, runtime build) before any status is emitted — even `connecting` only fires inside `serve_forever`. So a key rejected at the REST layer leaves the sampler at `Idle` forever, indistinguishable from "not started". Suggest the example keep the `JoinHandle` and surface its result, or have the early-failure path emit `closed`.

CHANGELOG.md
```diff
@@ -4,6 +4,34 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
...
+  attempts within one outage retain `reconnecting` without emitting duplicate
+  callbacks. Attempt-level detail remains available in lifecycle logs.
+- Python and TypeScript preserve make-before-break planned handoff and typed
+  bridge drain while making cold connection loss release blocked work promptly.
+  Rust retains its blocking, caller-owned thread model and cold reconnect path.
+- TypeScript `serveForever()` now rejects on terminal runtime failures, matching
+  `wait()`. Fire-and-forget callers must attach a rejection handler.
+- Python `close()` raises `TimeoutError` rather than reporting success if its
+  runtime thread cannot stop within the shutdown bound. Signal and
+  `KeyboardInterrupt` cleanup remain non-raising.
```
dimavrem22, line 35: Both of these are behavior breaks for 0.x consumers but are buried mid-`### Changed`. The 0.5.13/0.5.14 entries used `**Source-breaking:**` markers and a `### Compatibility` section — worth doing the same here: existing fire-and-forget `serveForever()` callers now risk an `unhandledRejection` crash, and `close()` in a `finally` can now raise and mask the original exception.

dimavrem22 (dismissed):
Solid direction — the bounded establishment/teardown, single cold-retry owner, and liveness surfaces are all things the tunnel runtimes needed, and this fixes several pre-existing issues along the way (TS leaked live sessions on failed attempts and orphaned ping-ack timers; a Rust stop-signal lost-wakeup race). Versions are in lockstep everywhere including the CLI dep range, and the docs/skills largely match the implementation.

Requesting changes for two confirmed bugs (Python: silent tunnel death on a failed handoff; Rust: unbounded dispatch-handle growth on healthy connections) and a non-compiling Rust example repeated across three docs — details inline.

### Cross-language consistency

Timeout values, backoff schedule, status vocabulary, terminal-error classification, and liveness semantics all line up across the three SDKs. Gaps found (inline comments where applicable):

- TS lacks the stale-reply generation fence that Python and Rust both gained in this PR.
- Cold-loss WS bridge close codes: Python now sends 1011 (or the planned drain codes), TS still sends a code-0 reset — customer-visible per-language difference.
- Python's `StatusCallback` type includes `"idle"`; TS explicitly excludes it via `Exclude<TunnelRuntimeStatus, "idle">`; the Rust docs list five states. `"idle"` is never emitted, so Python's callback type should probably exclude it too.
- Rust emits `connecting` even when `serve_forever` starts pre-stopped (`runtime.rs:334` fires before the stop check), where Python and TS emit only `closed`.
- `closed` emission ordering differs: TS emits at the top of `aclose`, Python at the end after teardown, and Rust only from inside `serve_forever` — so a Rust caller that never runs `serve_forever` gets no `closed` callback at all.

### Unit-test consistency

Each SDK currently proves scenarios the others skip. Since the changelog claims lockstep behavior, these should be leveled up so parity is actually pinned by tests:

| Scenario | Python | TypeScript | Rust |
|---|---|---|---|
| Backoff schedule + reset on success | **missing** | ✓ | ✓ |
| Edge-trigger status dedupe | **missing** | ✓ (exact status arrays) | ✓ |
| Status-callback exception isolation | **missing** | ✓ | ✓ |
| Cold-path HELLO timeout | ✓ | **only the handoff-budget variant** | ✓ |
| Auth failure during handoff is terminal | ✓ (but via a custom driver — see inline) | **code path untested** | n/a |
| Stale-reply / generation guard | ✓ (4 tests) | **mechanism absent** | **guard implemented but untested** |
| Handoff failure → cold reconnect | **missing** (and currently broken — see inline) | ✓ | n/a |

### Flakes

Several new Rust tests assert tight real-time margins (<100–150ms after `aclose`, and one exact-status-sequence test with a <320ms bound on a ~230ms nominal path) — inline comment with details. `start_paused` virtual time is already used elsewhere in the same file and would derisk these on loaded CI runners.

Also: no linked issue on the PR — if there's a tracking issue for the tunnel-recovery work, please link it.

sdk/python/inkbox/tunnels/client/_runtime.py
alex-w-99: Fixed cancelled-task result handling and added real-supervisor regressions for failed-handoff cold reconnect and clean shutdown. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/python/inkbox/tunnels/client/_runtime.py
alex-w-99: Fixed with compare-and-clear ownership, plus coverage for a second handoff starting during drain. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/python/inkbox/tunnels/client/_listener.py
alex-w-99: Documented the `TimeoutError` contract in the docstring, README, and skill. Kept the single-use 30-second bound local. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/typescript/src/tunnels/client/_runtime.ts
alex-w-99: Added TypeScript connection generations and stale-response rejection coverage to match Python and Rust. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/typescript/src/tunnels/client/_runtime.ts
alex-w-99: Fixed by suppressing reconnecting emissions after stop begins, with shutdown-race coverage. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/typescript/src/tunnels/client/_runtime.ts
alex-w-99: Aligned cold connection loss to a typed `WsConnectionLost` reset with code 1011 and added bridge coverage. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/rust/src/tunnels/client/runtime.rs
alex-w-99: Kept edge state recorded before invocation so a panicking callback cannot replay the same transition. Documented that callbacks must return promptly and replaced the misleading blocking-safety test. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/rust/src/tunnels/client/runtime.rs
alex-w-99: Pruned completed dispatch handles before each push and added a bounded-retention regression test. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/rust/README.md
alex-w-99: Updated the example to retain the `JoinHandle`, surface bootstrap/runtime errors, and explain why status remains `Idle` before runtime startup. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

README.md
alex-w-99: `Inkbox::new` returns `Arc<Inkbox>`, so this clone compiles. Updated the examples to retain and surface the tunnel thread result. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/typescript/src/tunnels/client/_listener.ts
alex-w-99: Attached the defensive handler when the promise is created while preserving rejection on the returned promise; added regression coverage. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

CHANGELOG.md
alex-w-99: Added a prominent Compatibility section covering both behavior changes and the required caller handling. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

sdk/rust/src/tunnels/client/runtime_tests.rs
alex-w-99: Changed the assertion to ordered-subsequence matching and relaxed real-time margins to reduce CI flake risk. [b44c14c](https://github.com/inkbox-ai/inkbox/commit/b44c14c)

dimavrem22 (approved):
Re-verified every thread against b44c14c8 — approving.

All of the earlier findings are properly addressed: the Python failed-handoff fix runs on the real supervisor path and its new regression tests fail with the fix reverted; the TS generation fence bumps only on cold teardown (handoff reply migration preserved) and covers all three route kinds; cold WS loss now reaches the customer close frame as 1011 in both Python and TS; the Rust dispatch-handle prune is on the hot path with a bounded-retention test; and the flaky timing assertions, docs, and changelog Compatibility section are all sorted. The notify-status pushback is fair — at-most-once edge delivery is the right call, and the new docs cover the promptness requirement.

Also conceding the `inkbox.clone()` comment: `Inkbox::new` returns `Arc<Inkbox>`, so the original examples compiled fine — my mistake.

Non-blocking leftovers, fine to ignore or fold into a later pass:

- Rust: `run_once` emits `connected` unconditionally after intake startup, so an `aclose` racing a mid-establishment connection can transiently observe `closed → connected → closed` (a `is_stopped()` gate before that emission would close it). Similarly, a takeover racing `aclose` can still flip a cleanly-closed status handle to `Superseded` (only the reverse direction is suppressed).
- TS: stale replies are dropped only after the handoff wait, so a stale reply can idle up to the 5s handoff budget before being discarded where Python drops it at entry — latency-only, no correctness impact.

PR #150, opened 2026-08-14
Add contact fact kinds, expiry, and hand-written facts

alex-w-99:
## Executive Summary

Adds contact-fact kinds and expiry to all three SDKs and the CLI, plus methods for writing and correcting a fact by hand.

- `kind` (`profile`, `preference`, `context`) and `expires_at` are optional on responses, so newer clients tolerate deployments that do not send them yet.
- New `contacts.facts.create()` and `update()` methods in Python, TypeScript, and Rust, plus matching CLI commands.
- Default lists omit expired context facts while retaining locked facts; include-expired options return expired facts too.

## Description

`ContactFact` gains optional `kind` and expiry fields: `Option<ContactFactKind>` / `Option<String>` in Rust, optional `kind` / `expiresAt` properties in TypeScript, and defaulted dataclass fields in Python. Response deserialization remains compatible when either field is absent.

Rust callers that construct `ContactFact` with struct literals must add `kind` and `expires_at`, usually as `None`. `serde(default)` preserves wire compatibility but cannot preserve source compatibility for existing literals. This remains release `0.5.17`.

`create()` and `update()` require an admin-scoped key. Hand-written facts never expire. Any update makes the fact user-authored, clears its expiry, and revives it if needed. Content changes additionally remove confidence and citations; kind-only changes retain them.

`create()` and `update()` enforce only the 10-fact limit for the kind being written. Facts Inkbox derives, and contact merges, are additionally held to 30 active facts per contact, with untyped facts counting toward that total. Merge errors name an overflowing kind or `total`; deleting any active fact can relieve total overflow.

## Testing

- `cd sdk/python && uv run pytest tests/test_contact_memory.py`
- `cd sdk/typescript && npm test -- tests/contactMemory.test.ts && npm run build`
- `cargo test --manifest-path sdk/rust/Cargo.toml`
- `cargo fmt --check --manifest-path sdk/rust/Cargo.toml`
- `cargo clippy --manifest-path sdk/rust/Cargo.toml --all-targets -- -D warnings`
- `cd cli && npm test`

Version remains `0.5.17` across all packages and the plugin manifest.


sdk/rust/src/contacts/types.rs
```diff
@@ -336,6 +357,12 @@ pub struct ContactFact {
     #[serde(default, deserialize_with = "f64_from_number_or_string")]
     pub confidence: Option<f64>,
     pub origin: ContactFactOrigin,
+    /// Memory bucket the fact belongs to, when it has been classified.
```
dimavrem22, line 360: codex: These optional fields are still source-breaking for Rust callers that construct `ContactFact` literals: an old initializer now fails E0063 for missing `kind` and `expires_at`; `serde(default)` only affects deserialization. Because `^0.5.16` admits 0.5.17, this patch update can break downstream builds. Preserve the old representation or release and document this at a source-breaking version boundary; also correct the PR body, which calls the string field `Option<DateTime>`.

CHANGELOG.md
```diff
@@ -4,6 +4,36 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
...
+- Create and edit facts by hand from Python, TypeScript, Rust, and the CLI:
+  `contacts.facts.create()` / `update()`, and `inkbox contacts facts create` /
+  `update`. Both require an admin-scoped API key; an agent-scoped key is
+  rejected with 403. Hand-written facts never expire, and the create call
+  returns 409 once a contact is at its limit for that kind.
+- `inkbox contacts facts list` shows `kind` and `expiresAt`.
+
+### Changed
+
+- Listing a contact's facts no longer returns context facts whose `expires_at`
```
dimavrem22: codex: `expires_at` is not unconditionally when a fact stops applying: a locked fact remains active and is returned by the default list after that timestamp. Consumers following this wording can locally hide memory that the API still treats as live. Document that `locked_at` overrides expiry across the changelog, SDK type/list docs, CLI, and bundled skills.

sdk/python/inkbox/contacts/resources/contacts.py
```diff
@@ -269,8 +269,10 @@ def merge(
     ) -> Contact:
         """Merge contacts into the contact identified by ``contact_id``.
 
-        Requires an admin-scoped API key. The server rejects the merge atomically
-        if the result would exceed 25 active memories; delete unwanted facts and retry.
+        Requires an admin-scoped API key. Active memories are budgeted per kind,
```
dimavrem22: codex: Merge also enforces a contact-wide total, so mixed typed and untyped facts can fail with a `total` error even when no individual kind is over. This guidance promises named kinds and gives no valid recovery for that case. Document both per-kind and total overflow across the SDKs, CLI, changelog, README, and skills; deleting any active fact can relieve total overflow.

sdk/python/inkbox/contacts/resources/contact_facts.py
```diff
@@ -20,15 +21,77 @@ class ContactFactsResource:
...
+
+    def update(
+        self,
+        contact_id: UUID | str,
+        fact_id: UUID | str,
+        *,
+        content: str | None = None,
+        kind: ContactFactKind | str | None = None,
+    ) -> ContactFact:
+        """Edit a fact's content or kind. Requires an admin-scoped API key; an
```
dimavrem22, line 76: codex: Update has a larger lifecycle effect than documented: any edit promotes a generated fact to user-authored, clears its expiry, and revives an expired fact. Correcting a context memory can therefore make it permanent unknowingly. Document these effects across all SDK and CLI update surfaces, and assert the returned origin/expiry in the SDK tests.

sdk/python/inkbox/contacts/resources/contacts.py
alex-w-99: Confirmed. Merge guidance now covers both per-kind and contact-wide total overflow across SDK docs, CLI help/README, changelog, and bundled skills. It explains that any active fact can relieve a `total` overflow. Updated in https://github.com/inkbox-ai/inkbox/commit/257da00.

sdk/python/inkbox/contacts/resources/contact_facts.py
alex-w-99: Confirmed. Every SDK and CLI update surface now documents that any edit makes the fact user-authored, clears expiry, and revives expired facts; content changes additionally remove confidence and citations. Python, TypeScript, and Rust tests assert returned origin/expiry. Updated in https://github.com/inkbox-ai/inkbox/commit/257da00.

sdk/rust/src/contacts/types.rs
alex-w-99: Confirmed. `serde(default)` only preserves deserialization compatibility; existing Rust struct literals must add the two fields. We are retaining `0.5.17`, and both the changelog and PR body now explicitly document that source break. The PR body also correctly uses `Option<String>` for `expires_at`. Updated in https://github.com/inkbox-ai/inkbox/commit/257da00.

CHANGELOG.md
alex-w-99: Confirmed. Type docs, list docs, CLI help/README, changelog, and bundled skills now state that expiry applies to unlocked generated context facts and that locked facts remain active in default lists. Updated in https://github.com/inkbox-ai/inkbox/commit/257da00.

sdk/python/inkbox/contacts/resources/contacts.py
dimavrem22: codex: Partially addressed. The changelog still says the limit applies per kind “rather than as one flat total” and that untyped memories have “their own allowance,” then immediately introduces a contact-wide total. Replace that opening with the actual contract: both per-kind and 30-total limits apply, and untyped memories count only toward the total.

alex-w-99: Fixed in [f9efbe7](https://github.com/inkbox-ai/inkbox/commit/f9efbe7). The changelog now states the complete contract: 10 active memories per named kind, 30 active memories total, and legacy untyped memories count only toward the total.

dimavrem22: codex: The contradiction is gone, but the replacement now overstates the 30-total rule. Manual fact creation intentionally enforces only the selected kind’s cap, so a contact with legacy untyped facts can exceed 30 via manual writes; qualify the 30-total ceiling to extraction/merge behavior rather than saying every contact can hold at most 30.

alex-w-99: Fixed in [f34d404](https://github.com/inkbox-ai/inkbox/commit/f34d404). The changelog now distinguishes the limits by write path: manual writes enforce the selected named kind’s 10-memory cap, while extraction and merge also enforce the 30-memory total, including untyped memories.

dimavrem22:
codex: ✅ Addressed in the rewritten head: the PR body and commit history now correctly distinguish kind-only manual writes from the 30-total extraction/merge behavior.

dimavrem22:
codex: ✅ Addressed in rewritten head `3f68442`: the attribution footer is removed and the code tree remains unchanged.

alex-w-99:
Both confirmed.

The PR body stated the per-kind and per-contact limits as a single rule. `create()` and `update()` are held only to the limit for the kind being written; the per-contact total applies to facts Inkbox derives and to contact merges. The body now says that.

`c225b3c` claimed the limit applies per kind rather than as a total and that only deleting facts of the named kinds frees room. Both halves are wrong: a total exists, and a merge rejected on the total is relieved by deleting any active fact. It is reworded in place as `18c5765`, with the three follow-ups rebased onto it as `eee3f77`, `470e991`, and `e2e0351`. The trees are unchanged. Commit links in the replies above point at pre-rebase SHAs.

alex-w-99:
Done, though the premise is only half right.

That trailer is not foreign to this repository: it appears on roughly half of `main`, including two commits in exactly the form used here. It has fallen out of use recently, which is the better argument, and the three sibling commits on this branch do not carry it, so the branch is more consistent without it.

Removed. `18c5765` is now `eb82de2`, with the follow-ups rebased to `11cbc96`, `7bdb605`, and `3f68442`. The tree is byte-identical across both rewrites; only messages and SHAs moved. Commit links earlier in this PR point at superseded SHAs.

If the trailer should be dropped repository-wide rather than on this branch alone, that is worth settling separately, since it would apply to a large share of existing history.

dimavrem22 (approved):
Approved based on the current code review. Before merging: mark this PR ready for review, let all required CI and integration workflows run to completion, and merge only after they are green and the documented rollout dependency/order is satisfied.

PR #151, opened 2026-08-17
Add email drafts across SDKs and CLI

alex-w-99:
"\"## Executive Summary\\n\\nAdds complete saved email draft support to the Python, TypeScript, and Rust SDKs and the CLI.\\n\\n- Covers idempotent create, list, get, update, duplicate, delete, send, and attachment operations.\\n- Preserves optimistic concurrency and explicit-null update semantics across languages.\\n- Ships documentation, agent skills, lifecycle coverage, and lockstep version `0.5.18`.\\n\\n## Description\\n\\nIntroduces first-class draft resources and mailbox-scoped identity helpers in every SDK. Draft models expose generation and send state, update inputs distinguish omitted fields from explicit null values, and attachment downloads preserve bytes, filename, and content type. Successful sends reuse the existing message models, while API and CLI errors retain stable error codes and `Retry-After` timing.\\n\\nAdds the `inkbox email drafts` command tree, including attachment add, remove, and download commands. Shared CLI attachment handling avoids changing existing email send behavior. Public documentation and skills explain generation conflicts, attachment index lifetimes, idempotent send retries, and uncertain delivery handling.\\n\\n## Reason\\n\\nAgents need to save incomplete email, resume edits across clients, and send only the exact revision they reviewed. Exposing the lifecycle consistently prevents callers from rebuilding draft and MIME behavior themselves.\\n\\n## Decisions\\n\\n- **Concurrency:** Require the latest draft generation on every mutation so concurrent mailbox or API edits cannot be silently overwritten.\\n- **Partial updates:** Represent omitted, explicit-null, and replacement values distinctly so callers can preserve or clear nullable fields intentionally.\\n- **Retry recovery:** Preserve structured conflict details and retry timing, and expose optional create idempotency keys so callers can safely retry creation or reconcile send outcomes.\\n- **Release:** Move all SDK, CLI, and plugin version surfaces together to `0.5.18`.\\n\\n## Testing\\n\\n- Python SDK focused draft tests: 24 passed; Ruff passed. The earlier full suite passed with 1058 tests and 3 skips.\\n- TypeScript SDK focused draft tests: 21 passed; build passed. The earlier full suite passed with 1063 tests and 3 skips.\\n- Rust SDK formatting and the full all-features suite passed with 327 tests; the final retry-metadata regression also passed.\\n- CLI `npm test` and `npm run build`: 127 tests passed; build passed.\\n- TypeScript SDK and CLI integration projects `npx tsc --noEmit`: lifecycle additions compile.\\n- Python lifecycle `pytest --collect-only`: draft lifecycle coverage collects successfully.\\n- Live integration lifecycle: Not run because deployed draft endpoints were not part of local validation.\\n\"\n"


cli/src/commands/drafts.ts
```diff
@@ -0,0 +1,480 @@
...
+    .option("--limit <n>", "Max drafts to show", positiveInteger, 50)
+    .action(withErrorHandler(async function (
+      this: Command,
+      options: { identity: string; limit: number },
+    ) {
+      const global = getGlobalOpts(this);
+      const inkbox = createClient(global);
+      const mailbox = await resolveMailbox(inkbox, options.identity);
+      const items: DraftSummary[] = [];
+      for await (const draft of inkbox.drafts.list(mailbox, { pageSize: options.limit })) {
```
dimavrem22: codex: `--limit` is described as the total drafts to show, but it is also passed through as the API page size. Any value above the server's 100-item page limit fails validation instead of paginating to the requested total. Cap `pageSize` at 100 while stopping after `options.limit` items, or validate and document this option as 1–100.

alex-w-99: Fixed in [b793edc](https://github.com/inkbox-ai/inkbox/commit/b793edc). The CLI now caps each API page at 100 while continuing pagination until the requested total `--limit` is reached. Added coverage for `--limit 101`.

dimavrem22 (approved):
codex: Reviewed the implementation across Python, TypeScript, Rust, and CLI. The draft lifecycle, generation checks, idempotent create, attachment operations, structured errors, and retry metadata align with the server contract. No blocking issues found.

PR #164, opened 2026-08-31
fix(tunnels/python): reset stream with upstream-error if ASGI handler raises after head

Adityakk9031:
## Summary

In `inkbox.tunnels.client._callable_streaming`, when an ASGI app streams a response in passthrough mode, `head_sent` is set to `True` once headers (`http.response.start`) are sent.

If the ASGI app subsequently raises an exception mid-stream (e.g., database connection loss, LLM stream interruption, network timeout, or unhandled generator exception), the `except Exception:` block caught the exception and logged it, but because `head_sent` was `True`, the `if not head_sent:` block was skipped. As a result, neither `response.reset("upstream-error")` nor `response.end_body()` was ever invoked.

This left downstream HTTP/2 streams hung indefinitely waiting for more data frames or a stream reset, leaking streams on the persistent HTTP/2 tunnel connection.

The TypeScript implementation (`sdk/typescript/src/tunnels/client/_callable_streaming.ts`) already handles this via `await response.reset("upstream-error")`.

## Proposed Fix

- In `invoke_asgi_streaming`, add an `else:` branch in the exception handler when `head_sent` is `True` to call `await response.reset("upstream-error")`.
- Added unit tests in `sdk/python/tests/test_tunnels_callable_streaming.py` verifying:
  1. Exception before head sends 502 with `upstream error`.
  2. Exception after head resets stream with `upstream-error`.

## Verification

```bash
pytest tests/test_tunnels_callable_streaming.py
# 6 passed in 0.80s
```

### DIFF
--- a/sdk/python/inkbox/tunnels/client/_callable_streaming.py
+++ b/sdk/python/inkbox/tunnels/client/_callable_streaming.py
@@ -145,3 +145,8 @@ async def invoke_asgi_streaming(
                 await response.end_body()
             except Exception:
                 pass
+        else:
+            try:
+                await response.reset("upstream-error")
+            except Exception:
+                pass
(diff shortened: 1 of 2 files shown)


Adityakk9031:
@rayruizhiliao have a look

PR #165, opened 2026-08-31
fix(vault): guard AES-GCM ciphertext length before buffer slicing

Adityakk9031:
## Summary

In both the TypeScript (`sdk/typescript/src/vault/crypto.ts`) and Python (`sdk/python/inkbox/vault/crypto.py`) SDKs, encrypted vault payloads are formatted as `ciphertext || nonce (12B) || tag (16B)`, requiring a minimum payload length of 28 bytes.

When decrypting a truncated or corrupted payload ($< 28$ bytes):
- In TypeScript: `blob.slice(-28, -16)` returns an empty `Uint8Array(0)` when the input length is shorter than 16 bytes. Passing a 0-byte IV to Node's `createDecipheriv("aes-256-gcm", key, nonce)` threw an unhandled engine error:
  ```
  TypeError [ERR_CRYPTO_INVALID_IV]: Invalid IV length 0 (must be 12 bytes)
  ```
- In Python: `blob[-28:-16]` returns a slice with length $< 12$ bytes, causing `AESGCM(key).decrypt(...)` to raise `ValueError: Nonce must be at least 12 bytes`.

In both languages, this resulted in uncaught low-level runtime errors rather than throwing `InkboxVaultKeyError`.

## Proposed Fix

- Added payload length validation before buffer slicing in `aesGcmDecrypt` (TS) and `_aes_gcm_decrypt` (Python):
  ```typescript
  if (blob.length < AES_IV_BYTES + AES_TAG_BYTES) {
    throw new InkboxVaultKeyError("Invalid ciphertext: payload is too short");
  }
  ```
- Added unit tests in `sdk/typescript/tests/vault/crypto.test.ts` and `sdk/python/tests/test_vault_crypto.py` asserting that truncated base64 payloads to `unwrapOrgKey` / `unwrap_org_key` and `decryptPayload` / `decrypt_payload` consistently reject with `InkboxVaultKeyError`.

## Verification

**TypeScript:**
```bash
npx vitest run tests/vault/crypto.test.ts
# 26 passed in 10.98s
```

**Python:**
```bash
pytest tests/test_vault_crypto.py
# 25 passed in 8.47s
```

### DIFF
--- a/sdk/python/inkbox/vault/crypto.py
+++ b/sdk/python/inkbox/vault/crypto.py
@@ -159,6 +159,8 @@ def _aes_gcm_decrypt(key: bytes, blob: bytes, aad: str = "") -> bytes:
         blob: Ciphertext blob.
         aad: Additional authenticated data (must match what was used during encryption).
     """
+    if len(blob) < AES_IV_BYTES + 16:
+        raise InkboxVaultKeyError("Invalid ciphertext: payload is too short")
     ct = blob[:-28]
     nonce = blob[-28:-16]
     tag = blob[-16:]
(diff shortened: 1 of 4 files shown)


Adityakk9031:
@rayruizhiliao have a look

PR #166, opened 2026-09-01
Optional delivery auth token on webhook subscriptions

dimavrem22:
## Executive Summary

Webhook subscriptions can now carry an optional bearer token that is attached to every delivery; reads return the stored token (`auth_token`/`authToken`), and the CLI's `--auth-token-stdin` keeps the secret out of argv and shell history.

- `auth_token` on create, tri-state `auth_token` on update (absent = unchanged, `null` = clear, string = replace) across the Python, TypeScript, and Rust SDKs and the CLI.
- Every delivery and replay then carries `Authorization: Bearer <token>` in addition to the existing `X-Inkbox-*` signature headers.
- Reads expose only a boolean `has_auth_token` / `hasAuthToken` (defaults to false on older servers); no read ever returns the token.

## Description

Adds the delivery auth token to all four packages. Python's `WebhookSubscriptionsResource.create` gains a keyword-only `auth_token` (omitted from the body when `None`) and `update` uses the shared `_UNSET` sentinel so an explicit `None` passes through as JSON `null` to clear the stored token. TypeScript mirrors this with `authToken?: string` on create options and `authToken?: string | null` on update options, mapped to the `auth_token` wire key exactly like the existing `contextConfig` handling. Rust extends `create` with a trailing `auth_token: Option<&str>` and `update` with tri-state `auth_token: Option<Option<&str>>`, following the `context_config` precedent. The read model gains `has_auth_token: bool` (`#[serde(default)]` in Rust, `?? false` in TS, `d.get(..., False)` in Python) so payloads from servers that predate the field keep parsing.

The CLI adds `--auth-token` to `webhook subscription create` and `--auth-token` / `--clear-auth-token` (mutually exclusive, erroring if combined) to `webhook subscription update`, and surfaces `hasAuthToken` in subscription list/get/create output. READMEs, the three bundled skills, and the root plus package changelogs are updated, and all package versions move in lockstep to 0.6.6.

## Reason

Some webhook endpoints sit behind gateways that require their own `Authorization` header, which the signature headers alone cannot satisfy. This lets a subscription authenticate to such endpoints; the stored token is readable back from subscription reads. Requires server-side support.

## Decisions

- **Version 0.6.6, not 0.6.4**: main is already at 0.6.5 (and 0.6.4 is an existing release), so this ships as the next patch bump with lockstep versions per RELEASING.md.
- **Tri-state update mirrors `context_config`**: the token reuses the established omit/null/replace pattern (Python `_UNSET` sentinel, TS `undefined` vs `null`, Rust `Option<Option<...>>`) rather than introducing a new convention, and the Python sentinel is imported from `inkbox.identities.types` so identity checks compare against one object.
- **No client-side token validation**: the server owns the 1–4096-char single-line constraint; the SDKs pass the string through so the limit lives in one place, consistent with how the exact event-name enum is server-validated.
- **Breaking surface documented, not avoided**: TS `WebhookSubscription` object literals must add `hasAuthToken`, and the Rust `create`/`update` signatures grow a trailing argument — both follow the precedent set by the `context_config`/forwarding additions and are called out in the root CHANGELOG's Compatibility section.

## Testing

- Python: `webhooks.subscriptions.create(..., auth_token="your-endpoint-token")` posts the key and the response parses `has_auth_token=True`; `create` without the kwarg omits the key; `update(id, auth_token="...")` patches the string, `update(id, auth_token=None)` patches `null`, and an unrelated update omits the key; parsing defaults a missing `has_auth_token` to `False`. Suite: `uv run pytest` — 1088 passed, 3 skipped.
- TypeScript: the same six cases via exact-body `toHaveBeenCalledWith` assertions in `tests/webhooks-subscriptions.test.ts`. Suite: `npx vitest run` — 1091 passed, 3 skipped; `npx tsc --noEmit` clean.
- Rust: serde test covers `has_auth_token` default-false and explicit-true parsing. Suite: `cargo test` — 256 passed + doc-tests; `cargo clippy --all-targets --features tunnels-runtime -- -D warnings` and `cargo fmt --check` clean.
- CLI: `flattenCreateForOutput` exposes `hasAuthToken` (and never a token field). Suite: `npm test` (build + `node --test`) — 129 passed. Manual: `inkbox webhook subscription create --mailbox-id <id> --url <url> --event-type message.received --auth-token your-endpoint-token` then `update <sub-id> --clear-auth-token`; passing both `--auth-token` and `--clear-auth-token` errors.


alex-w-99 (approved):
Reviewed the current head. No blocking concerns remain.

PR #167, opened 2026-09-02
Make Voice AI model selection platform-managed

dimavrem22:
## Executive Summary

Voice AI model selection is now platform-managed while existing SDK and CLI calls remain source-compatible.

- Python, TypeScript, and Rust retain their model arguments as documented no-ops.
- The CLI retains `--model` as a documented no-op for script compatibility.
- Every release surface receives a patch version bump.

## Description

The SDKs no longer send a model override when replacing Voice AI configuration. Existing method signatures, response fields, and parsers remain intact, allowing applications to work across rollout order and continue parsing older responses.

Current guidance and bundled skills omit model configuration. Core Python, TypeScript, Rust, CLI, and Claude package versions move from `0.6.6` to `0.6.7`; the independently versioned Codex and Cursor plugin manifests move to `0.1.2` and `1.0.1`.

## Reason

Voice AI has one supported platform selection, so accepting a model argument as active configuration was misleading. Compatibility requires retiring its behavior without breaking existing source code or scripts.

## Decisions

- **API compatibility:** Kept all public method arguments and the CLI flag as no-ops instead of removing them in a breaking release.
- **Response compatibility:** Retained nullable model fields and tolerant parsers so clients work with responses from either side of the rollout.
- **Versioning:** Applied patch bumps to every shipped package and plugin manifest because all affected surfaces remain backwards compatible.

## Testing

- Python `set_hosted_agent_config(model=...)`: Expected the call to succeed without sending the model; 1,089 tests passed with 3 skipped.
- TypeScript `setHostedAgentConfig({ model: ... })`: Expected the call to succeed without sending the model; build passed and 1,091 tests passed with 3 skipped.
- Rust `set_hosted_agent_config(..., model, ...)`: Expected the call to succeed without sending the model; 256 tests and doctests passed, with Clippy and rustfmt clean.
- CLI `inkbox phone hosted-agent set --model ...`: Expected the flag to remain accepted and documented as ignored; 130 tests passed.
- Version parity: Expected core release surfaces to report `0.6.7` and independent plugin manifests to contain their patch releases; manifest parsing and lockstep tests passed.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "inkbox",
   "description": "Inkbox communication tools and SDK skills for Claude Code.",
-  "version": "0.6.6",
+  "version": "0.6.7",
   "author": {
     "name": "Inkbox AI",
     "email": "hello@inkbox.ai"
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "inkbox",
-  "version": "0.1.1",
+  "version": "0.1.2",
   "description": "Inkbox communication tools and SDK skills.",
   "author": {
     "name": "Inkbox AI",
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "inkbox",
   "displayName": "Inkbox",
-  "version": "1.0.0",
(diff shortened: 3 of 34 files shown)


alex-w-99 (approved):
Reviewed 5af3eb854 across Python, TypeScript, Rust, CLI, versions, changelogs, and public surfaces. Deprecated model inputs remain source-compatible no-ops, while response fields remain compatible. No blocking concerns.

PR #172, opened 2026-09-06
`0.6.9`: Document contact create saving a matching suggestion

alex-w-99:
## Executive Summary

Document that creating a contact saves a matching suggested contact, and prepare release 0.6.9.

- Add guidance beside contact creation examples in the Python, TypeScript, and CLI skills.
- Bump SDKs, CLI, and the bundled plugin to 0.6.9, with the Codex manifest at 0.1.4.
- Preserve the voice-discovery functionality and 0.6.8 changelog from main; no contact client implementation changes.

## Description

When request identifiers match one eligible unreviewed contact, the API confirms that contact with the submitted profile, keeps its memories and existing identifiers, and returns it. Saved contacts, multiple matching contacts, and conflicted matching identifiers still return HTTP 409 `duplicate_contact_identifier`. The existing clients already support the response fields, so this PR updates guidance and release metadata only.

Merged `origin/main` at `eb1ab59a`. Resolved the changelog conflict by retaining the 0.6.8 voice-discovery entry and placing contact guidance under 0.6.9. Updated package manifests, lockfiles, the CLI SDK dependency, both User-Agent constants, and the bundled skill manifests together. The three skill merges retain both voice-discovery and contact guidance.

Review follow-up `f83e8b85` explicitly documents agent-scoped create support and distinguishes replacement names from preserved omitted non-name fields. Existing SDK and CLI request builders already omit unsupplied profile fields; no client implementation changes are needed.

## Reason

Agents need accurate guidance when adding an address that already belongs to a suggested contact. A new manifest version makes the updated skills available to existing installations.

## Decisions

- **Next patch release:** main already uses 0.6.8 and Codex 0.1.3, so use 0.6.9 and 0.1.4 rather than reuse those release versions.
- **No client behavior changes:** existing contact methods already accept and return the required fields.
- **Cursor manifest unchanged from main:** it ships no skills, so retain its existing 1.0.2 version.

## Testing

- Review-follow-up validation: Python contact/memory tests 35 passed; TypeScript contact/memory/version tests 22 passed; Rust contact tests 13 passed; CLI contact/version tests 12 passed. Python lock and diff checks passed. The full-suite/build results below were run after the main merge; the follow-up changes only guidance.
- Validation target: create a contact matching one eligible suggestion through the SDKs or `inkbox contacts create --json`; expect the existing ID, confirmed review status, and retained memories once the API behavior is available. A saved-contact match must still return HTTP 409. Live API validation was not run.
- Python: `uv lock --check`, `uv run pytest -q --tb=short` (1,095 passed, 3 skipped), and `uv run ruff check .` passed.
- TypeScript: `npm run build`, `npx vitest run --reporter=dot` (1,094 passed, 3 skipped), and `npm run verify:bundle` passed.
- CLI: `npm test` passed all 134 tests, including release-version consistency and voice-catalog commands, after rebuilding the TypeScript SDK. The previously reported stale-build type errors no longer reproduce.
- Rust: `cargo test --locked -q` passed 259 tests; `cargo test --locked --features tunnels-runtime` passed 342 tests. Each run also passed 3 documentation tests with 1 ignored. Clippy with warnings denied and `cargo fmt --check` passed.
- `git diff --check` passed. No packages published; publish the TypeScript SDK before the CLI per `RELEASING.md`.


dimavrem22 (commented):
Checked the lockstep bump against `RELEASING.md` (no `0.6.7` left outside the changelog; CLI `@inkbox/sdk` dep, both User-Agent constants, lockfiles, Rust, and the Claude/Codex manifests all move) and read the three skill paragraphs. No blockers.

One dependency: the skill text and changelog say the call "saves a matching suggested contact" unconditionally. I have asked for the server-side change to require the same authority the API already requires for confirming a contact (a console session or an admin API key; agent-scoped keys receive 403 when they send `review_status`). If that lands, agent-scoped keys keep getting 409 and these three paragraphs need "when called with an admin API key" so agents reading the skill do not expect a save that will not happen. Hold the publish until the server-side decision is settled.

alex-w-99:
@dimavrem22 The create behavior remains available to agent-scoped API keys, so an admin-only qualifier would be inaccurate. Explicit confirmation through a contact update remains a separate operation with its existing permissions. No packages have been published.

[f83e8b85](https://github.com/inkbox-ai/inkbox/commit/f83e8b85) clarifies all three skills and the changelog: creation can save a matching suggestion with an agent-scoped key, replaces its name fields, preserves omitted non-name profile fields, and applies supplied non-name fields. Saved-contact, multiple-owner, and conflicted-identifier matches still fail.

The version review predates the main merge: this PR now targets 0.6.9 with Codex 0.1.4, retaining the 0.6.8 release entry. The release-consistency test passes. Contact-focused validation passed: Python 35, TypeScript 22, Rust 13, CLI/contact-version checks 12; Python lock and diff checks also passed.

dimavrem22 (approved):
Re-reviewed `f83e8b85`. The bump now targets 0.6.9 with Codex 0.1.4 after 0.6.8 shipped on `main`; every file in `RELEASING.md` moves, the CLI `@inkbox/sdk` dependency and both User-Agent constants match, and the 0.6.8 release entry is retained. The three skill paragraphs and the changelog describe the shipped behavior accurately: agent-scoped keys can save a suggestion, name fields are replaced, omitted non-name fields are preserved, and saved-contact, multi-owner, and conflicted matches still 409. That matches what each SDK sends on the wire, since all of them omit unset scalars. No blockers.

PR #174, opened 2026-09-07
Standardize scheduled integration notifications

alex-w-99:
## Executive Summary

Standardize scheduled integration failure notifications.

- Show environment, status, and test type on separate lines.
- Keep component and package source in labeled details.
- Separate notification threads by run, attempt, and environment.

## Description

Updates the Python, TypeScript, and CLI workflow notifications to use the same three-line header and Summary, Scope, and Details fields. Each environment has a distinct emoji, with no brackets around its name. Rendering tests exercise the workflow shell with synthetic inputs and run in CI. The notification step fails visibly when the webhook post is rejected or the environment label is unknown, instead of posting an unlabeled message or exiting quietly.

## Reason

Consistent formatting makes failures easier to scan and keeps unrelated runs out of the same thread.

## Decisions

- Share a run/attempt/environment thread across failing components; component and package source remain visible in each message.
- Keep existing alert conditions and scheduled test behavior unchanged; no package release is needed for workflow-only changes.

## Testing

- `node --test .github/tests/notification-format.test.mjs`: 24 passed. Renders each notification variant, checks header order and fields, validates thread identity using synthetic values, and confirms an unknown environment fails before anything is posted.
- No live notifications were sent.

### DIFF
--- a/.github/tests/notification-format.test.mjs
+++ b/.github/tests/notification-format.test.mjs
@@ -1,0 +1,53 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import { spawnSync } from 'node:child_process';
+import { test } from 'node:test';
+
+const workflow = readFileSync(new URL('../workflows/sdk_integration.yml', import.meta.url), 'utf8');
+const components = { 'python-sdk': 'Python SDK', 'typescript-sdk': 'TypeScript SDK', cli: 'CLI' };
+const title = '📦 SDK integration tests';
+const environments = { development: '🛠️ Development', beta: '🧪 Beta', production: '🚀 Production' };
+
+// Execute the actual workflow shell with a local curl replacement: no network or secrets.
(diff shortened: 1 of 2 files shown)


dimavrem22 (approved):
Approving. Reviewed all three workflow notification steps and the rendering tests (24 passed locally). Untrusted values go through `jq --arg`, the unknown-environment guard fails before anything is posted, and the thread identity is consistent across the Python, TypeScript, and CLI paths. No blockers.

Two non-blocking notes:

1. `notification-format-tests` runs inside `sdk_integration.yml`, which also fires on `schedule`. A failing format test on a scheduled run would flip the whole run to failure and read as an integration outage in the alerts space. Gating the job with `if: github.event_name != 'schedule'` keeps harness problems out of scheduled results.
2. The notify step keeps `--retry-all-errors` on a POST without a client message ID, and `--fail-with-body` now makes 4xx responses count as failures, so a transient 4xx can post the root twice. Pre-existing shape, just slightly widened.

PR #176, opened 2026-09-08
`0.6.10`: Add release-connection methods to the SDKs and CLI

alex-w-99:
## Summary

Agents could list their active iMessage connections but had no way to end one; only the recipient could, by texting the router. This release adds a per-connection release to every package, calling the new `DELETE /api/v1/imessage/assignments/{assignment_id}`.

- **Python:** `imessages.release_assignment(id)` and `identity.release_imessage_assignment(id)`
- **TypeScript:** `imessages.releaseAssignment(id)` and `identity.releaseIMessageAssignment(id)`
- **Rust:** `imessages().release_assignment(&id)` and `identity.release_imessage_assignment(&id)`
- **CLI:** `inkbox imessage disconnect <assignment-id>` (no identity flag: the release is organization-wide)

Behavior: inbound from the recipient stops routing to the agent and the line can be reassigned. The recipient is not notified and can reconnect by texting the router again. Requires an admin API key; identity-scoped keys are rejected. A missing or already-released connection is a not-found error.

Version bumped 0.6.9 to 0.6.10 in lockstep (all four packages, both User-Agent constants, lockfiles). Bundled plugin manifests move with it: Claude `0.6.10`, Codex `0.1.5`, Cursor `1.0.3`. Changelog entry added. Skills and READMEs carry a one-line example.

## Testing

- Python: `uv run pytest` in `sdk/python`, 1096 passed. New test asserts `DELETE /assignments/{id}`.
- TypeScript: `npx vitest run` in `sdk/typescript`, 1095 passed. New test asserts method, URL, and empty body.
- Rust: `cargo test imessage`, 27 passed; `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean. New mock-server test for the DELETE.
- CLI: `npm test`, 134 passed; `tsc --noEmit` clean against the rebuilt SDK.
- Not exercised against a live environment; the server side ships separately.

https://claude.ai/code/session_01MD6cENFdt9SisoQ98UG2LL

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "inkbox",
   "description": "Inkbox communication tools and SDK skills for Claude Code.",
-  "version": "0.6.9",
+  "version": "0.6.10",
   "author": {
     "name": "Inkbox AI",
     "email": "hello@inkbox.ai"
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "inkbox",
-  "version": "0.1.4",
+  "version": "0.1.5",
   "description": "Inkbox communication tools and SDK skills.",
   "author": {
     "name": "Inkbox AI",
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "inkbox",
   "displayName": "Inkbox",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "description": "Give Cursor an Inkbox identity for email, SMS, iMessage, voice calls, contacts, notes, and agent-to-agent tasks.",
   "author": {
     "name": "Inkbox",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
 Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
 (Python), `@inkbox/cli`, `inkbox` (Rust, crates.io), and the bundled plugin.
 
+## 0.6.10 — Release an iMessage connection
+
+### Added
+
+- `imessages.release_assignment` / `identity.release_imessage_assignment`
+  (Python), `imessages.releaseAssignment` /
+  `identity.releaseIMessageAssignment` (TypeScript), `release_assignment` /
+  `release_imessage_assignment` (Rust), and `inkbox imessage disconnect
+  <assignment-id> -i <handle>` (CLI) call the new
+  `DELETE /imessage/assignments/{assignment_id}`. Releasing a connection stops
+  routing the recipient's inbound to the agent and frees the shared line. The
+  recipient is not notified and can reconnect by texting the triage number.
+  Identity-scoped keys can only release their own identity's connections; a
+  missing or already-released connection is a not-found error.
+
 ## 0.6.9 — Creating a contact saves a matching suggestion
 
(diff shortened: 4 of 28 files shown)


dimavrem22 (changes requested):
One blocker, otherwise this looks right.

**Blocker — the bundled plugin manifests are not bumped in lockstep.** This PR changes `skills/inkbox-cli/SKILL.md`, `skills/inkbox-python/SKILL.md`, and `skills/inkbox-ts/SKILL.md`, and `.codex-plugin/plugin.json` bundles `./skills/`, so the Codex plugin ships changed content at an unchanged `0.1.4`. Every prior release in `CHANGELOG.md` bumped the manifests alongside the packages (0.6.7 → Codex `0.1.2` / Cursor `1.0.1`, 0.6.8 → `0.1.3` / `1.0.2`, 0.6.9 → `0.1.4`), and the 0.6.9 commit touched all three manifest files. Please bump `.codex-plugin/plugin.json` to `0.1.5` (and `.cursor-plugin/plugin.json` to `1.0.3` if you want to keep the previous pattern — it bundles no skills, so that one is a judgement call) and add the usual "Bundled plugin manifests receive patch releases: …" line to the 0.6.10 entry.

Everything else checked out:
- Python, TypeScript, and Rust resource methods all hit `DELETE /assignments/{id}` through the existing no-body `delete` transport helpers, and the identity-level wrappers reuse the existing iMessage-enabled guards (`_require_imessage` / `_requireIMessage` / `require_imessage`).
- Package versions, `CLI_VERSION`, `VERSION`, the CLI's `@inkbox/sdk` range, and all four lockfiles are at `0.6.10` with no `0.6.9` leftovers outside the changelog history.
- Docstrings, READMEs, and skills describe the same behavior (admin key only, recipient not notified, reconnect via the router, 404 on missing/released).

Non-blocking: `inkbox imessage disconnect` takes `-i <handle>`, but the identity is only used for the iMessage-enabled check — the release itself is organization-wide, so the command will release an assignment that belongs to a different identity than the one passed. Fine as-is given the admin-key requirement, just worth knowing.

Ran locally on this head: Python SDK `pytest` → 1096 passed, 3 skipped; TypeScript SDK `vitest` → 1095 passed, 3 skipped; CLI `npm test` (typecheck against the rebuilt 0.6.10 SDK + 134 tests) → clean; Rust `cargo test imessage` → 27 passed.

alex-w-99:
Addressed in 8f046ae2b81adaa0d53667cb61c94aa991058fe0:

- Codex manifest bumped to `0.1.5` and Cursor to `1.0.3` (kept the prior pattern), with the "Bundled plugin manifests receive patch releases" line added to the 0.6.10 changelog entry.
- Took the non-blocking note as well: `inkbox imessage disconnect` no longer takes `-i`. The release is organization-wide, so the flag implied a scope the command never enforced. It now calls `inkbox.imessages.releaseAssignment` directly. Skill and changelog updated to match.

CLI typecheck and `npm test` (134) clean on this head.

dimavrem22 (approved):
Re-reviewed at 8f046ae. The blocker is resolved: Codex `0.1.5` and Cursor `1.0.3` are bumped and the 0.6.10 changelog entry now carries the manifest line. Dropping `-i` from `inkbox imessage disconnect` is the right call — the command goes through the public `inkbox.imessages` accessor the triage command already uses, and the skill/changelog text is updated to match with no leftover references to the flag.

Re-ran CLI `npm test` on this head against the rebuilt 0.6.10 SDK: typecheck clean, 134 passed. SDK sources are unchanged since the earlier run (Python 1096, TypeScript 1095, Rust 27 all passing). Approving.

mathematicalobsession:
Epic.🔥🔥🔥🔥🔥

PR #180, opened 2026-09-10
0.7.0: Add contact access and communication permission APIs

alex-w-99:
## Executive Summary

Add contact visibility and per-address communication controls across Python, TypeScript, Rust, and the CLI in release `0.7.0`.

- Read and partially update whole-group email/phone visibility independently from permission to contact each address.
- Create contacts with initial access atomically, and expose access settings in the administrative contact roster.
- Retain boolean permissions, guarded advanced policies, filtered previews, and compatibility access listings.

## Description

Adds `contacts.access.get/update` in Python and TypeScript, `contacts().access().get/update` in Rust, and `contacts access get/set` in the CLI. Requests support partial `email`/`phone` objects with `visible` and `contactable`, plus Profile/Memories booleans. Serialization preserves false values, empty lists, empty group objects, and omitted fields. `ContactCreatePermissions` supports nested group choices or the existing boolean maps, and roster entries expose optional `access` settings separately from communication coverage.

The PR also includes boolean permission methods, advanced per-address policy editing with revision/expected-action checks, identity previews, paginated management rosters, and nullable matching contact cards on communication rules. Phone choices cover SMS, calls, and iMessage together. CLI file validation, public exports, synthetic transport fixtures, skill guidance, and the canonical changelog are aligned with the prepared lockstep `0.7.0` release.

## Reason

Developers need to share contact information without authorizing communication to every visible address, with consistent SDK types and CLI commands that preserve partial-update intent.

## Decisions

- **Access API:** Extend the existing access resource with selected-agent get/update methods while retaining its legacy read-only `list` method, avoiding a second competing resource name.
- **Partial updates:** Preserve omitted fields instead of supplying defaults because visibility-only edits must not replace communication choices; an explicit empty `contactable` list blocks every current address in that group.
- **Profile parent:** Reject outbound Email, Phone, or Memories grants when the resulting Profile permission is disabled; response parsing remains tolerant during rollout.
- **Visibility:** Document whole-group visibility independently from exact-address communication; inherited visibility can expose additional addresses in an existing mixed group without authorizing contact to them.
- **Creation:** Extend `ContactCreatePermissions` with optional nested groups and preserve the boolean-map shape; callers choose one address representation per request.
- **Compatibility:** Preserve absent/null roster access and existing wire methods. Rust **274 passed**


dimavrem22 (commented):
Reviewed 0.6.11 against the backend contract. No blockers: field names, casing, enums, nullability, the `visibility` block, the roster shape, and pagination match the API in all three SDKs and the CLI, the phone-rule guard removal mirrors the server-side change, and the version bump is complete (no `0.6.10` leftovers in any manifest, lockfile, user agent, or plugin manifest). Public-surface grep is clean, and this PR actually removes a few older leaks from docstrings and fixtures. Ran the Python suite locally on 3.11: 897 passed (the 17 tunnel collection errors are the pre-existing missing `h2` dependency).

Inline comments below are should-fix / nit level. Two things that don't anchor:

- **Skill trees miss the new identity helper.** `list_contact_communication_policies` / `listContactCommunicationPolicies` exist on `AgentIdentity` in all three SDKs but are absent from the `AgentIdentity` trees in `skills/inkbox-python/SKILL.md` (~L55-62) and `skills/inkbox-ts/SKILL.md` (~L54-61).
- **Tests are self-consistent mocks.** Every fixture hand-codes the snake_case names, so the suites would still pass if the server renamed `full_profile`, `has_more`, or `expected_revision`. The TS and Rust preview fixtures also omit `visibility` entirely even though the server always sends it; only the Python preview test uses the real shape. Not a blocker, but worth one recorded-response fixture per language for the policy endpoints.

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
...
+- The selected mode determines active entries: allows in whitelist mode and blocks in blacklist mode. Opposite-action entries no longer override the active list.
+- Agent contact responses, lookup, and vCard exports include only permitted identifiers and profile fields. Profile and Memories are independently controlled; inherited settings preserve the existing full-view requirement. Hosted voice in YOLO mode retains full organization contact and memory access.
+- **Authorization change:** communication-rule creation now requires admin credentials, like updates and deletion. Agent keys receive HTTP 403. Existing-contact identifier edits and suggestion absorption also require administrative authority.
+- Identity-level phone-rule helpers no longer require a dedicated phone number. Existing phone/iMessage methods remain supported; their filter-mode fields are aliases and contradictory values return HTTP 422.
+- Releasing or replacing a phone number preserves identity permissions. Migration preserves a sole configured channel's restrictions and ignores never-used/unconfigured defaults.
```
dimavrem22: The second sentence ("Migration preserves a sole configured channel's restrictions and ignores never-used/unconfigured defaults") describes a server-side data migration, not anything an SDK caller can observe or act on. Same for the "Hosted voice in YOLO mode retains full organization contact and memory access" sentence on L22. Suggest trimming the Changed section to client-visible behavior; the rollout mechanics belong in the private coordination surface.

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
 Versions move in lockstep across `@inkbox/sdk` (TypeScript), `inkbox`
 (Python), `@inkbox/cli`, `inkbox` (Rust, crates.io), and the bundled plugin.
 
+## 0.6.11 — Contact communication permissions
+
+### Added
+
+- Contact communication-policy get/replace, identity previews, and paginated identity views in Python, TypeScript, and Rust. The CLI adds `contacts communication-policy get`, `set`, `preview`, and `list`.
```
dimavrem22: Nit: the CLI list here omits `list-management` (mentioned only on L15) and `identity contact-policies`, both of which ship in this release. Worth listing them together so the Added section is the single place someone scans.

sdk/typescript/src/contacts/resources/communicationPolicy.ts
```diff
@@ -0,0 +1,166 @@
...
+  identities: ContactIdentityVisibilityDecisions[];
+}
+/** Effective access, independent of whether the group has stored data. */
+export interface ContactVisibilityResult { profile: boolean; memories: boolean }
+export interface ContactCommunicationPolicy {
+  contactId: string;
+  revision: number;
+  defaults: ContactChannelDecisions;
+  identities: ContactIdentityDecisions[];
+  visibility?: ContactVisibilityPolicy;
```
dimavrem22: Question: `visibility` is optional on the policy **response** here, and likewise `ContactVisibilityPolicy | None` / `ContactVisibilityResult | None` in Python and `Option<_>` in Rust. The server marks `visibility` as required on both the policy response and the preview, and the endpoints are brand new, so there is no older response shape to tolerate. Every caller now has to write `policy.visibility?.defaults` for no reason. Was this deliberate? (Optional on the *request* is correct and matches omit-to-preserve.)

sdk/rust/src/contacts/resources/communication_policy.rs
```diff
@@ -0,0 +1,511 @@
...
+#[derive(Debug, Clone, Deserialize)]
+pub struct ContactVisibilityResult {
+    pub profile: bool,
+    pub memories: bool,
+}
+
+/// Complete contact communication settings and optimistic revision.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ContactCommunicationPolicy {
+    pub contact_id: String,
```
dimavrem22: ID typing is inconsistent inside this module: `contact_id: String` here, `identity_id: String` on L32/L47/L96, but `id: uuid::Uuid` on L127 and the existing rule/identity types use `Uuid` throughout. The server sends UUIDs for all of them. The tests lean on the `String` typing (`"contact-1"`, `"agent"` fixtures), so switching means updating fixtures, but it is cheaper now than after 0.6.11 is on crates.io.

cli/src/commands/contacts.ts
```diff
@@ -31,7 +32,90 @@ function collectValues(value: string, previous: string[] = []): string[] {
...
+        ...row.effective, revision: row.revision })),
+      { json: !!opts.json, columns: ["id", "contact", "email", "phone", "profile", "memories", "revision"] });
+    }));
+  policy.command("get <contact-id>").description("Read a contact policy (admin credentials)")
+    .action(withErrorHandler(async function (this: Command, contactId: string): Promise<void> {
+      const opts = getGlobalOpts(this);
+      output(await createClient(opts).contacts.communicationPolicy.get(contactId) as unknown as Record<string, unknown>, { json: !!opts.json });
+    }));
+  policy.command("set <contact-id>").description("Replace a contact policy (admin credentials)")
+    .requiredOption("--file <path>", "JSON file with expectedRevision, defaults, and identities")
```
dimavrem22: Nit: help text says the file carries `expectedRevision, defaults, and identities`, but `parseContactPolicyFile` also accepts the optional `visibility` block (and the skill docs mention it). Add it here so `--help` matches what the command validates.

cli/src/commands/contacts.ts
```diff
@@ -31,7 +32,90 @@ function collectValues(value: string, previous: string[] = []): string[] {
...
+    .option("--q <query>", "Search contact details")
+    .option("--limit <number>", "Page size", "50").option("--offset <number>", "Page offset", "0")
+    .addOption(new Option("--order <order>", "Sort order").choices(["name", "recent"]).default("recent"))
+    .addOption(new Option("--review-status <status...>", "Contact review states").choices(["confirmed", "unreviewed"]))
+    .action(withErrorHandler(async function (this: Command, handle: string, options: {
+      q?: string; limit: string; offset: string; order: "name" | "recent"; reviewStatus?: ContactReviewStatus[];
+    }): Promise<void> {
+      const opts = getGlobalOpts(this);
+      const page = await createClient(opts).contacts.communicationPolicy.listManagementForIdentity(handle, {
+        ...options, limit: Number(options.limit), offset: Number(options.offset),
```
dimavrem22: Nit: `Number(options.limit)` on `--limit abc` yields `NaN`, which the HTTP layer still serializes, so the user gets a server 422 instead of a local argument error. The existing rule commands use `parseInt` with a check; same here and on the `list` command below (L114).

CHANGELOG.md
alex-w-99: Removed the migration-specific sentence and kept the observable behavior that permissions survive number replacement. Retained the YOLO sentence deliberately: YOLO is an exposed voice mode, and its organization-wide contact/memory read behavior is important for callers to distinguish from ordinary scoped reads.

Commit: [96b68858](https://github.com/inkbox-ai/inkbox/commit/96b68858).

CHANGELOG.md
alex-w-99: Fixed: the first Added entry lists get, set, preview, list, list-management, and identity contact-policies together.

Commit: [96b68858](https://github.com/inkbox-ai/inkbox/commit/96b68858).

sdk/typescript/src/contacts/resources/communicationPolicy.ts
alex-w-99: Agreed. Policy and preview response visibility is now required in Python, TypeScript, and Rust; request visibility remains optional with omit-to-preserve semantics. Updated fixtures and transport tests accordingly. These interfaces are still part of the unpublished 0.6.11 release.

Commit: [96b68858](https://github.com/inkbox-ai/inkbox/commit/96b68858).

sdk/rust/src/contacts/resources/communication_policy.rs
alex-w-99: Fixed: contact_id and identity_id in the Rust policy/override/preview models now use uuid::Uuid. Updated fixtures to valid UUIDs, adjusted callers, and documented the struct-literal requirement in the root changelog. Rust tests and Clippy pass.

Commit: [96b68858](https://github.com/inkbox-ai/inkbox/commit/96b68858).

cli/src/commands/contacts.ts
alex-w-99: Fixed: set --help now includes optional visibility alongside expectedRevision, defaults, and identities; the CLI help regression checks it.

Commit: [96b68858](https://github.com/inkbox-ai/inkbox/commit/96b68858).

cli/src/commands/contacts.ts
alex-w-99: Fixed for both contact-policy list commands and identity contact-policies. Validation rejects malformed/fractional values and enforces limit 1-200 and offset 0-10000 before HTTP. Command-level tests verify invalid inputs produce a local error and zero requests, rather than serializing NaN or accepting parseInt prefixes.

Commit: [96b68858](https://github.com/inkbox-ai/inkbox/commit/96b68858).

alex-w-99:
Review fixes are pushed in [96b68858](https://github.com/inkbox-ai/inkbox/commit/96b68858), with replies on each inline.

Added the identity policy-list helpers to both language skill trees. The shared synthetic policy/update/preview fixture now runs through each SDK's HTTP serialization and was separately validated, losslessly, against the current API models. This improves the contract check without presenting synthetic data as a recorded live response; it does not replace live compatibility testing. All policy/preview fixtures now include required visibility.

The h2 dependency is already declared as h2>=4.1 in sdk/python/pyproject.toml, so no dependency change was needed. A complete environment passes 1,115 Python tests (3 skipped). TypeScript: 1,113 passed (3 skipped). Rust: 350 unit tests and 3 doctests passed, formatting/Clippy passed. CLI: 140 tests and build passed.

The root CHANGELOG.md remains the canonical release log. Version 0.6.11 remains unpublished; its description now includes the full four-group/roster contract and actual verification results.

CHANGELOG.md
```diff
@@ -4,6 +4,41 @@ All notable changes to the Inkbox SDK, CLI, and skills live here.
...
+- Access PATCH preserves omitted choices; `contactable: []` blocks all current addresses, and `visible: true` alone preserves communication. Hiding Profile hides omitted or empty email/phone groups and omitted Memories; explicit choices in the same request win. Hiding a group also blocks identifiers subsequently added to it; an explicit exact-address allow reveals a uniquely matched group without allowing its other addresses.
+- Rust `ContactCreatePermissions` struct literals must include `email` and `phone` or use `..Default::default()`. `ContactPermissionEntry` literals must include `access`, using `None` when absent. Older roster responses still parse.
```
dimavrem22: **LOW.** `ContactCreatePermissions`, `ContactPermissionEntry`, and `communication_policy.rs` are new in this PR, so "struct literals must include `email`/`phone`" and "IDs now use `uuid::Uuid`" (L39) read as migrations from an API that never shipped. Also, L22-44 mostly describe backend behavior changes rather than package changes — worth a separate "Behavior notes" heading so package consumers can tell them apart.

sdk/python/inkbox/contacts/types.py
```diff
@@ -48,6 +48,73 @@ class ContactNameSource(StrEnum):
     IDENTIFIER_FALLBACK = "identifier_fallback"
 
 
+@dataclass
```
dimavrem22, line 51: **LOW.** `ContactCreatePermissions` (same in `types.rs:13-27` and `contacts.ts:361-366`) accepts the `emails`/`phones` maps together with the `email`/`phone` group objects. The API treats `permissions` as a union of three strict shapes, so a mixed body fails every branch with an unhelpful 422. A client-side check (or a test pinning the 422) would make the changelog's "do not mix" note enforceable.

cli/src/commands/identity.ts
```diff
@@ -522,7 +536,7 @@ export function registerIdentityCommands(program: Command): void {
     .option("--contact-sharing-enabled <bool>", "Toggle automatic name and optional photo sharing on a dedicated iMessage line: true or false")
     .option("--imessage-filter-mode <mode>", "iMessage contact-rule mode: whitelist or blacklist (admin-only)")
```
dimavrem22: **LOW.** Help text still says "iMessage contact-rule mode … (admin-only)"; it is now an alias of the phone filter mode and conflicting values are rejected. L539 was updated, this one was not.

skills/inkbox-ts/SKILL.md
```diff
@@ -42,7 +42,7 @@ Inkbox (admin-only client)
 ├── .mailContactRules         → MailContactRulesResource    (DEPRECATED — per-mailbox)
 ├── .phoneContactRules        → PhoneContactRulesResource   (DEPRECATED — per-number)
 ├── .smsOptIns                → SmsOptInsResource
-├── .contacts                 → ContactsResource   (.facts, .correspondence, .access, .vcards)
+├── .contacts                 → ContactsResource   (.communicationPolicy, .facts, .correspondence, .access, .vcards)
```
dimavrem22: **LOW.** Tree omits `.permissions` (`ContactsResource.permissions` exists at `contacts.ts:128`; the Python tree lists it).

dimavrem22 (commented):
## External review pass 3 — head `744e54c2` (2026-09-14)

Two independent review tracks (Claude + Codex), merged. Verdict: **close** — request/response shapes, omitted-vs-null handling, paths, pagination, exports, and the tolerant parser all check out against the API contract, and the Python (1119 passed) and TypeScript (1118 passed, `tsc` clean) suites run green from the worktree. Remaining items are packaging and docs polish.

### Should fix
- Bundled Codex/Cursor plugin manifests not bumped although all four `skills/*/SKILL.md` changed (the 0.6.10 release bumped both and listed it in the changelog).
- `ContactCreatePermissions` lets callers mix the `emails`/`phones` maps with the `email`/`phone` group objects; the API rejects a mixed body with a confusing 422 and nothing client-side enforces the "do not mix" note (`CHANGELOG.md:36`).

### Polish
- `cli/README.md:617-642` still documents only `contacts access list`; none of `access get/set`, `permissions get/set`, `communication-policy *`, `identity contact-policies` appear.
- Version is `0.7.0` rather than the usual `+0.0.1` (`0.6.11`). Consistent across all 13 version sites and defensible given the Rust struct-literal changes — author's call, just flagging the departure.
- Rust closed enums (`ContactDecision`, `ContactAddressKind`, `IdentifierPermission`) have no unknown-variant fallback; matches crate style, hardening only.
- Python `communication_policy.py` does not `quote()` handle/contact_id, unlike `contact_access.py` and `permissions.py`.
- Rust compile/clippy and the CLI `node --test` (imports unbuilt `dist/`) were not run here — static review only for those.

### Checked and clean
All three SDKs + CLI: `ContactAccessSettings`, `ContactPermissions`, `ContactCommunicationPolicy`, preview, roster page shapes and enums; PATCH serializers never emit `null` and omit unset groups; `contactable: []` sent literally; PUT omits `visibility` when unset; CLI file parsers reject `null`/string booleans/unknown keys; every changelog-named symbol and every name in the four SKILL.md files exists at head; fixture `tests/fixtures/contact_communication_policy.json` matches the response models. No sensitive or internal detail introduced anywhere in the diff; several stale internal references are removed.


### Findings outside diff hunks (could not anchor inline)
- `.codex-plugin/plugin.json:3` — **MED.** All four `skills/*/SKILL.md` changed and this manifest points at `./skills/`, but only the Claude plugin manifest was bumped. The 0.6.10 release (8f046ae2) bumped Codex and Cursor for exactly this reason and listed it in the changelog; same needed here (and `.cursor-plugin/plugin.json:4`).

sdk/python/inkbox/contacts/types.py
alex-w-99: Fixed in all three SDKs. TypeScript makes the two shapes mutually exclusive and validates at runtime; Python and Rust reject mixed shapes before HTTP. Added request-prevention regressions for each runtime path. Commit: [2ef8a640](https://github.com/inkbox-ai/inkbox/commit/2ef8a64055d7b2ba35c226ef323432b557e7b385).

skills/inkbox-ts/SKILL.md
alex-w-99: Fixed: the TypeScript skill tree now includes `contacts.permissions`. Commit: [2ef8a640](https://github.com/inkbox-ai/inkbox/commit/2ef8a64055d7b2ba35c226ef323432b557e7b385).

cli/src/commands/identity.ts
alex-w-99: Fixed: the help text now identifies this as the shared phone/iMessage mode alias and matches the admin API-key requirement. Commit: [2ef8a640](https://github.com/inkbox-ai/inkbox/commit/2ef8a64055d7b2ba35c226ef323432b557e7b385).

CHANGELOG.md
alex-w-99: Fixed the misleading migration claims for types introduced by this unreleased change: removed the UUID migration note and reframed the new Rust fields as additions. Retained consumer-visible behavior in Changed because it affects how these SDK methods resolve access. Commit: [2ef8a640](https://github.com/inkbox-ai/inkbox/commit/2ef8a64055d7b2ba35c226ef323432b557e7b385).

dimavrem22 (approved):
## Pass 4 — head `2ef8a640` — approved

Re-verified: version aligned to `0.6.11` across Python, TypeScript, Rust, CLI, and the bundled Claude/Codex manifests (Cursor is MCP-URL based, no bump needed); `ContactCreatePermissions` now rejects mixed shapes in all three SDKs with regressions; CLI README documents every new command; identity help text and the TypeScript skill tree are corrected; changelog no longer describes migrations for unreleased types.

Shapes, omitted-vs-null serialization, paths, pagination, exports, and the tolerant parser were verified in pass 3 and are unchanged. Python and TypeScript suites green from this checkout. Publish only after the backend change is live.

PR #183, opened 2026-09-14
`0.6.11`: Add idempotency keys to email send, reply-all and forward

alex-w-99:
## Summary

Email sends, reply-alls and forwards now accept an optional idempotency key, so a caller whose response is lost or times out can retry without delivering the email twice.

| Package | Surface |
|---|---|
| Python | `idempotency_key=` on `messages.send` / `reply_all` / `forward`, and on `identity.send_email` / `reply_all_email` / `forward_email` |
| TypeScript | `idempotencyKey` in the options object of the same methods |
| Rust | new `send_with_idempotency_key`, `reply_all_with_idempotency_key`, `forward_with_idempotency_key` on `MessagesResource`, and the three `*_email_with_idempotency_key` equivalents on `AgentIdentity` |
| CLI | `--idempotency-key <key>` on `inkbox email send`, `reply-all` and `forward` |

Each reuses the existing 1–255 character key validator and attaches the header only when a key is supplied, so a call without one produces the exact same request as before.

## Semantics

This is at-most-once, not a replay. A repeat under a key that already sent raises rather than returning the original message: `409` when the email was accepted, `503` when an earlier attempt's outcome is unresolved.

- Keys are scoped per organization and per method, so one key value used for a send and for a forward does not collide.
- A key lasts 7 days, after which it can be used again.
- The request body is not part of the key, so reuse it only for a retry of the same email, never for a different one.
- An attempt that errors still consumes its key. Use a new key rather than retrying a failed send under the old one.

## Rust stays source-compatible

`MessagesResource::send`, `reply_all` and `forward` take positional arguments, so adding the key to them would have changed their arity and broken every existing caller on a routine `cargo update`. Instead each keeps its current signature and delegates to a private implementation, and the keyed form is a sibling method. The `*_with_idempotency_key` naming matches `post_with_headers` and `post_with_timeout` already in the crate.

Every published signature in all four packages still compiles unchanged: the diff on the Rust sources adds six `pub fn` lines and removes none, and every pre-existing test and doctest call site is untouched.

## Version

All four packages and the plugin manifest move to **0.6.11**, with the lockfiles, both User-Agent constants and the CLI's `@inkbox/sdk` dependency range updated, plus a `CHANGELOG.md` entry.

## Testing

- `cd sdk/python && pytest` — 1096 passed, 3 skipped.
- `cd sdk/typescript && npx vitest run` — 1095 passed, 3 skipped; `tsc --noEmit` clean.
- `cargo test --manifest-path sdk/rust/Cargo.toml` — 262 unit tests and 3 doctests pass; `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.
- `cd cli && npm test` — 134 passed; `tsc --noEmit` clean.

New Rust coverage asserts that the keyed path sends the `Idempotency-Key` header and rejects an empty key before any request leaves the client, and that the unkeyed path sends no such header.

To exercise it end to end: send twice with one key and confirm the second call is refused and only one email arrives; then send twice with no key and confirm both are delivered.

---
<a href="https://app.blacksmith.sh/inkbox-ai/codesmith/inkbox/pr/183"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792005703&installation_model_id=436612&pr_number=183&repository=inkbox-ai%2Finkbox&return_to=https%3A%2F%2Fgithub.com%2Finkbox-ai%2Finkbox%2Fpull%2F183&signature=06454d7748551aedb897989c4722a8233c5a47af9270fa0e566e48ac2ab07929"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


dimavrem22 (approved):
codex: No code blockers in the SDK/CLI implementation. Verified optional header propagation across all four languages/surfaces and preserved Rust signatures. Locally, 59 focused Python tests and 92 TypeScript tests passed; additionally checked all three Python methods for keyed headers, omitted headers, and empty-key rejection. Rust CI is green. Release remains dependent on the backend idempotency fix being live.

dimavrem22 (approved):
codex: Re-reviewed the latest commit. Retry guidance now consistently preserves the same key after uncertain failures across Python, TypeScript, Rust, and the changelog. Runtime changes remain as previously reviewed, and current SDK CI is green. No code blockers found. Publish only once the matching API behavior is live.

